PR #6, opened 2012-02-06
changes for icob due to new structure

fmessmer:
please verify this version and let me know whether the structure is ok!


fmessmer:
Since the pull request is closed now, i assume you found the relevant commits. It only were the most recent commits dating Jan/Feb 2012. Now, is there anything I still need to do?

PR #34, opened 2012-08-23
Some small config changes for real robot.

abubeck:


cob_bringup/components/ur5.launch
```diff
@@ -2,7 +2,8 @@
 <launch>
 
  	<!-- universal robot arm -->
-        <node name="ur5_driver" ns="arm_controller" pkg="ur5_driver" type="driver.py" args="ur5" output="screen">
+        <node name="ur5_driver" ns="arm_controller" pkg="ur5_driver" type="driver.py" args="192.168.41.40" output="screen">
```
floweisshardt, line 5: this should go into a configuration file in cob_hardware_config. Is there no way to set a parameter for that instead of args?

cob_bringup/robots/raw3-1.launch
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0"?>
```
floweisshardt, line 1: please merge to ipa320 first, some things have changed regarding arg and env handling

abubeck, line 1: merged yesterday. Still changes?

cob_bringup/components/ur5.launch
abubeck, line 5: Not in the current version of universal_robot. We can submit a patch but have to handle it like this for now.

abubeck:
Fixed strange merging problems.

cob_hardware_config/raw3-1/config/light.yaml
```diff
@@ -1,3 +1,3 @@
-devicestring: /dev/ttyUSB2
+devicestring: /dev/ttyUSB0
```
floweisshardt, line 1: info: there is a udev rule in https://github.com/ipa320/setup/blob/master/udev_rules/01-cob.rules which maps the LED board to ttyLed

PR #37, opened 2012-08-24
new arm_joint_configurations

fmessmer:


fmessmer:
ToDo:
- get this position for all other robots as well
- send e-mail with hint to use this arm_position for mounting/dismounting the sdh to 32x-cob3*

PR #40, opened 2012-10-05
respawn base_velocity_smoother due to frequent crashes, see Trac-Ticket ...

mgruhler:
...#354

temporarily "fixes" the problem by respawning the node every time it crashes.

Might lead to small disturbances in the trajectory or stuttering, but does not stop the robot all the time.


floweisshardt:
that's no solution: I won't accept this pull request

PR #41, opened 2012-10-19
hardware monitor added

uhr-eh:


floweisshardt:
please add a dependency to cob_hwboard in the manifest.xml

you can check dependency problems with executing a "make test" in cob_bringup

PR #47, opened 2012-11-21
added inversion flag param to light configuration

benmaidel:
this is needed for the cob_light driver. raw3s lightboard inputs are inverted!


floweisshardt:
please use "false" as default value for "invert_output". That way we can keep all cob yaml files as they are and only add the parameter to the raw yaml files. please revert the changes to the cob yaml files.

benmaidel:
removed inversion flag from cob yaml files!

PR #55, opened 2013-01-25
Updated manual and hztests in cob_hardware_test

ipa-fmw-ms:
Removed old single hztest of kinect, lasers, prosilica and the all_in_one's
Added tests for calibration, joy and teleop 
Added hztests for testing all topics in one bringup with yaml configuration
Added rviz config for calibration
Added hztests yaml configuration for desire and cob3-3
Replaced desire and cob3-3 auto.xml tests with robot_auto.test
Calibration works only local in a local simulation not with jenkins, robot and a separate rviz-machine


floweisshardt:
- please keep old testing files to be able to start a single test (e.g. only test kinect)
- please separate cob_hardware_test/config/hztests_cob3-3.yaml into one yaml file per component to be able to use it in single tests

PR #60, opened 2013-03-07
Modifications for raw3-3 with Frida and fixes inside light_controller

benmaidel:


cob_bringup/components/light.launch
```diff
@@ -3,15 +3,9 @@
 
 	<arg name="robot" default="$(optenv ROBOT !!NO_ROBOT_SET!!)"/>
 	<arg name="pkg_hardware_config" default="$(find cob_hardware_config)"/>
-	<arg name="sim" default="false" />
```
NadiaHG, line 6: why did you delete cob_light_sim?

the simulated cob_light node should publish the marker topic but not communicate to the hardware device

benmaidel, line 6: Since there is no cob_light_sim node, we do not need this launch configuration anymore. The new cob_light package enables simulation mode automatically if no hardware is present. Publishing the LightMarker can be enabled by setting a Flag inside light.yaml configuration.

PR #83, opened 2013-05-27
modifications and fixes for raw's

benmaidel:


fmessmer:
As I merged with @ipa-bnm before doing my changes, this pull request should be obsolete. The changes will be included in pull request #84 as well.

PR #129, opened 2013-11-06
updated cob3-1 config

frederikhegger:


floweisshardt:
please revert changes to cob_controller_configuration_gazebo/ros/controller/torso_v0_controller.yaml and cob_controller_configuration_gazebo/ros/launch/torso_v0_controller.launch because this will affect other robots too.

If you thing these changes are still usefull, please commit them to a dedicated feature branch and send a pull request only for that features. Then we can discuss there if it will fit to all other robot too.

frederikhegger:
But the configuration right now does not work. The controller_manager (ros_control) used in the launch file, can not work with controllers implementing the pr2_controller_interface (which the robot_mechansim_controllers do). So we either can change the controller back to pr2_controller_manager (but Alexander Bubeck already change this to controller_manager) or take our new configuration (the YAML files) which works.

floweisshardt:
hydro_dev branches from ipa320 are now tested in simulation

PR #131, opened 2013-11-15
Optimize control

fmessmer:
This pull request does some restructuring of the configs for the gazebo controllers.
I also tried to optimize the controller values.
For most of the robots and components I now get "success" reports in the result of the follow_joint_trajectory actions. 
Still needs to be refined a bit more - but maybe we give it another run after fixing the dynamic model (inertias,...) in the URDF


floweisshardt:
visual review: ok

floweisshardt:
tested on robot: OK

PR #149, opened 2014-03-13
Refactored phidget launch

benmaidel:
renamed phidget.launch to tray_sensos.launch and added launch and config files for the real phidget driver


floweisshardt:
visual review: OK

PR #169, opened 2014-04-04
tested on cob4-1

NadiaHG:


floweisshardt:
visual review: OK
tested on cob4-1

PR #173, opened 2014-04-09
Feature/extended monitoring

abubeck:


cob_monitoring/src/emergency_stop_monitor.py
```diff
@@ -16,9 +19,61 @@
 
 class emergency_stop_monitor():
 	def __init__(self):
-		rospy.Subscriber("/emergency_stop_state", EmergencyStopState, self.emergency_callback)
-		self.em_status = EmergencyStopState()
-		self.first_time = True
+		self.color = "None"
+		rospy.wait_for_service("/light_controller/mode")
```
floweisshardt: why do we need to wait for the mode service? This will not work (block forever) on robot not having a light controller and e.g. only having speech. This should at least be configurable (enable_light: true/false, enable_sound: true/false, ...)

abubeck: We need to wait for the led driver before sending led settings. Otherwise, after bringup, the led's are just off until the first state change appears instead of showing the initial state of the robot.
I agree with the enabling parameters. I will add that to the PR.

abubeck:
Added enabling and disabling of led and sound itf to PR.

floweisshardt:
visual review: OK

PR #188, opened 2014-05-21
Adapt cob_image_flip and new tag for openni2 driver

NadiaHG:


NadiaHG:
Tested on cob4-1: OK

the device_id for openni2 is  not working, #1 means first device found and #2 second device found

We should define the device_id using bus and device X@Y , but it doesn't work for the actual openni2 version

floweisshardt:
visual review: OK

PR #194, opened 2014-06-17
Fixed reestructuration errors

NadiaHG:


fmessmer:
Looks good, thank you!
Just two comments (I know they are minor, but I just like to have it consitent):
- call the base_driver.launch file before! the base_controller.launch
- the base_controller.launch file itself needs to be moved to cob_bringup/controllers folder (then again the launchfiles need to slightly be updated)

floweisshardt:
visual review :ok:

PR #209, opened 2014-07-23
Inverted scanners

NadiaHG:


floweisshardt:
visual review: OK

fmessmer:
3.1416 vs. 3.1415 vs. $(M_PI)?

cob_hardware_config/cob4-2/urdf/calibration_default.urdf.xacro
```diff
@@ -34,7 +34,7 @@
 	<!-- torso mount positions | relative to base_link -->
 	<property name="def_torso_x" value="0.0"/>
 	<property name="def_torso_y" value="0.0"/>
-	<property name="def_torso_z" value="0.33411"/>
+	<property name="def_torso_z" value="0.26411"/>
```
fmessmer, line 37: when using 0.26411 there are collision contacts between base and torso (you can see them in Gazebo when toggeling in 'VIEW')
Is this number from CAD? Can we set it a little larger? 
I don't know what the consequences are but I don't want them to effect the controller parameter tuning...

NadiaHG, line 37: This number is from CAD, we should use it because of the kinematic.
I suggest create a new collisions mesh file

NadiaHG:
xacro doesnt support the sum: ${def___x}+${offset___x} if the value one of those properties is also a property (M_PI) 

Error:  
result = float(result) + float(n)
ValueError: could not convert string to float: ${M_PI}

floweisshardt:
visual review: OK

PR #222, opened 2014-09-10
install tag

NadiaHG:


fmessmer:
Can you please also add the install tag for the cob_controller_adapter_gazebo.py in cob_controller_configuration_gazebo/src?
I guess we don't need iinstall tags for the test scripts in the scripts  folder, so that's ok...

floweisshardt:
I vote for having install tags for the test scripts too. Otherwise they should be deleted.

fmessmer:
Well, from just looking at the code they look okay. If they run i.e. can be executed without failure, we can add install tahs for them, too

fmessmer:
+1

I guess we could try to simplify the fake_diagnostics a bit by having a node that accepts a list of topics rather than a separate node per topic....it would really reduce the launch file...
But that would be done in a different PR...

floweisshardt:
visual review: OK

PR #226, opened 2014-10-07
bringup tests

NadiaHG:
*Added dependency cob_safety_controller 
*Removes arg <machine..> https://github.com/ros-drivers/openni2_launch/commit/bbcf9f5451c2536736537e4d35bb1d1175065090

### DIFF
--- a/cob_bringup/package.xml
+++ b/cob_bringup/package.xml
@@ -28,6 +28,7 @@
   <run_depend>ur_driver</run_depend>
   <run_depend>cob_relayboard</run_depend>
   <run_depend>cob_voltage_control</run_depend>
+  <run_depend>cob_safety_controller</run_depend>
   <run_depend>cob_sick_s300</run_depend>
   <run_depend>cob_sick_lms1xx</run_depend>
   <run_depend>cob_hokuyo</run_depend>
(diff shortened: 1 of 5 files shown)


cob_bringup/package.xml
```diff
@@ -28,6 +28,7 @@
   <run_depend>ur_driver</run_depend>
   <run_depend>cob_relayboard</run_depend>
   <run_depend>cob_voltage_control</run_depend>
+  <run_depend>cob_safety_controller</run_depend>
```
floweisshardt, line 31: we need to be careful here because cob_safety_controller is in a private repo.
so either we need to have an entry in cob_substitute or make the cob_safety_controller public. I vote for having it in cob_substitute.

NadiaHG, line 31: It is already in cob_substitute: https://github.com/ipa320/cob_substitute/tree/indigo_dev/cob_safety_controller

fmessmer, line 31: There already is a `cob_safety_controller` package in `cob_substitute` (since hydro)
see [here](https://github.com/ipa320/cob_substitute/tree/hydro_dev/cob_safety_controller)

fmessmer, line 31: should refresh tabs more often...;-)

floweisshardt:
perfect!

PR #230, opened 2014-10-22
Adds light configuration for cob4-2

thiagodefreitas:
@ipa-nhg  Please Review


cob_hardware_config/cob4-2/config/light.yaml
```diff
@@ -0,0 +1,8 @@
+devicestring: /dev/ttyUSB0
```
floweisshardt, line 1: @ipa-nhg: we should have an udev rule here to avoid confusion with other USB devices

floweisshardt:
opened issue in setup_cob4, see https://github.com/ipa320/setup_cob4/issues/8

PR #233, opened 2014-10-28
[Indigo] add parameter for max_X_velocity to launch file

### LINKED ISSUE #207: Add parameter to cob_hardware_config
https://github.com/ipa320/cob_driver/pull/148 introduced new parameters
- max_trans_velocity
- max_rot_velocity

Currently they are not set within any yaml file, thus the hasParam-check fails and the default value is used. It would be good to know in which yaml these parameters can be set (in cob_hardware_config?). There should be a respective entry in the according yaml for all robots (using the same default value as in the code)
--> reduces output when starting bringup/bringup_sim

@ipa-frm

fmessmer:
fixes #207 
Adds the parameters to `cob_bringup/drivers/base_driver.launch` file as suggested in #207 
This way, it solves the problem for both simulation and real hardware...

@ipa-frm @ipa-mig

### DIFF
--- a/cob_bringup/drivers/base_driver.launch
+++ b/cob_bringup/drivers/base_driver.launch
@@ -11,6 +11,8 @@
 		<param name="IniDirectory" value="$(arg pkg_hardware_config)/$(arg robot)/config/base/"/>
 		<param name="timeout" value="0.2"/>
 		<param name="PublishEffort" value="false"/>
+		<param name="max_trans_velocity" value="1.1"/>	<!-- default: 1.1 m/s -->
+		<param name="max_rot_velocity" value="1.8"/>	<!-- default: 1.8 rad/s -->
 
 		<!-- start underlying components - base_drive_chain -->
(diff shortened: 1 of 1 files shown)


mgruhler:
Visual review: okay

thanks @ipa-fxm for fixing this...

fmirus:
visual review :ok_hand:

floweisshardt:
this parameters should go into a robot specific yaml file in cob_hardware_config, because this will probably not be the same limits for all robots.

mgruhler:
@ipa-fmw you're right. But as this node currently still has the ini-file structure, I would do this once we got rid of the ini files. Espacially since those two parameters are only the absolute upper limit for avoiding the "racing robot bug"

floweisshardt:
I added it to https://github.com/ipa320/cob_robots/issues/207.
visual review: OK

ipa-fxm:
Well as #207 has been merge for a while now and no complains occurred about any robot driving too fast/slow...I think it is ok just leaving those default values...and consider this fixed...
no more effort for putting the parameter from launch file to yaml!

PR #234, opened 2014-10-28
[Indigo] uses forward command controller for all simulated bases

fmessmer:
Merge together with [cob_common#131](https://github.com/ipa320/cob_common/pull/131)

This PR improves simulated Odometry with a factor of about 10...especially when a "Zero-Twist" is sent, i.e. base should not move at all.

@ipa-mig and others from the Nav-Team please review (do you have students working with simulation?)


mgruhler:
This seems to make sense.
There are currently several students working with simulation.

Get's my Okay

floweisshardt:
visual review: OK

PR #237, opened 2014-10-30
[indigo] Delete unsupported robots

floweisshardt:


NadiaHG:
Visual review : OK

fmessmer:
Do we still need the ur_connector stuff when cob3-7 is removed?
This would mean removing it from `cob_controller_configuration_gazebo/controller` as well as from `cob_common/cob_description`

BTW, `cob_common/desire_description` can completely be removed!

floweisshardt:
I would still keep it, to be able to get it added again easily if needed. The point of deleting robots is just, that we do not have to test them

PR #255, opened 2014-12-16
set interpolation perdiod to sync interval (10ms/100Hz)

mathias-luedtke:
Elmo period was too low (-> buffer underrun), Schunk period is quite high.
Needs to be tested on real hardware (Schunk+Elmo).
Maybe we should increase it to relax jitter problems, but this may lead to problems with full buffers.


floweisshardt:
visual review: OK
tested on cob4-2: OK

PR #260, opened 2015-01-16
fix cartesian controller parameters for arms

floweisshardt:


floweisshardt:
visual review: OK
tested on cob4-2: OK

PR #267, opened 2015-02-10
Cob4 updates

NadiaHG:
- make base work for cob4-1 and cob4-2
- parameter updates for cartesian controller
- remove obsolete files

PR #266, opened 2015-02-02
setup cob46

NadiaHG:

PR #269, opened 2015-02-11
Adjustments on the base configuration files

thiagodefreitas:
- Joint Limits
- HW mode
- mode_adapter now optional
- dcf_overlay
- Homing speed parameters


thiagodefreitas:
Should come after https://github.com/ipa320/cob_robots/pull/266 and https://github.com/ipa320/cob_robots/pull/267

fmessmer:
Is it correct, that the commits related to cob4-6 only introduce the driver part (402) for the real hardware? I can't see any base_controller in the `cob_bringup/robots/cob4-6.xml`.
Simulation also doesn't seem to be set up for cob4-6, right?

I would like to review this PR in more detail in order to see whether it's working in simulation, too...

cob_bringup/drivers/canopen_402.launch
```diff
@@ -5,19 +5,21 @@
...
+	<node ns="$(arg component_name)" name="joint_states_relay" pkg="topic_tools" type="relay" args="joint_states /joint_states" cwd="node" respawn="true" output="screen"/>
+
 
     <!-- controllers -->
 	<rosparam ns="$(arg component_name)" command="load" file="$(arg pkg_hardware_config)/$(arg robot)/config/$(arg component_name)_controller.yaml" />
 	<node ns="$(arg component_name)" name="$(arg component_name)_joint_state_controller_spawner" pkg="controller_manager" type="spawner" args="joint_state_controller" respawn="true" />  
   
 	<!-- control_mode_adapter -->
-	<node ns="$(arg component_name)" name="control_mode_adapter" pkg="cob_control_mode_adapter" type="cob_control_mode_adapter_node" cwd="node" respawn="false" output="screen"/>
+	<node if="$(arg mode_adapter)" ns="$(arg component_name)" name="control_mode_adapter" pkg="cob_control_mode_adapter" type="cob_control_mode_adapter_node" cwd="node" respawn="false" output="screen"/>
```
fmessmer, line 23: Isn't this the driver launch file?
I think there should not be anything related to controllers, controller spawning and mode adaptation in it...

fmessmer, line 23: see my comment [here](https://github.com/ipa320/cob_robots/pull/270#issuecomment-74378750)
We should discuss were to put the joint_mode_adapter and we should handle the case when some of the control modes are not supported...

thiagodefreitas:
@ipa-fmw update: corrected suffixes

PR #274, opened 2015-03-12
[indigo] add led_components parameter to emergency_stop monitor

floweisshardt:
this is to support multiple led controllers on cob4 and remove the hardcoded light_controller namespace. Requires an additional parameter in the emergency_stop_monitor.yaml file which lists the led components.

Should be compatible with raw3-1 and raw3-3.

needs to be merged together with https://github.com/ipa320/cob_command_tools/pull/82


NadiaHG:
Visual Review: OK

PR #280, opened 2015-03-19
add mimic support for sss

floweisshardt:


NadiaHG:
Visual Review: OK

cob_default_robot_config/cob4-2/mimic_configurations.yaml
```diff
@@ -0,0 +1 @@
+service_ns: /set_mimic
```
floweisshardt, line 1: this is a full service name. we should move mimic into a standalone namespace and the use this as `service_ns`

PR #289, opened 2015-03-27
added cob4-4

NadiaHG:


NadiaHG:
Please merge together with https://github.com/ipa320/cob_calibration_data/pull/79

cob_hardware_config/cob4-4/config/Elmo.dcf
```diff
@@ -0,0 +1,4502 @@
+[FileInfo]
```
floweisshardt, line 1: can't we use the dcf file in the common directory? is there something special about cob4-4? if not, please use a softlink instead of copying

NadiaHG, line 1: @ipa-mdl the Elmo.dcf file is the same for all robots/components?

mathias-luedtke, line 1: Should be merged for now. We have to unify all configs to a final one later.

floweisshardt:
@ipa-fxm: still missing your critical feedback

fmessmer:
As it follows the structure of cob4-6, **OK** from my side!

PR #294, opened 2015-04-22
remap diagnostics for cob_head_axis

fmessmer:
This is related to https://github.com/ipa320/cob_driver/pull/215
As discussed there, an appropriate remap is more desirable than a hard-coded namespace

@jobstvogt FYI


floweisshardt:
visual review: OK

PR #299, opened 2015-05-11
cob_hardware_config: added laser-calibration data for raw3-3

fmirus:
needs ipa320/cob_common#156 and ipa320/cob_calibration_data#80

### DIFF
--- a/cob_hardware_config/raw3-3/urdf/calibration_default.urdf.xacro
+++ b/cob_hardware_config/raw3-3/urdf/calibration_default.urdf.xacro
@@ -87,4 +87,13 @@
 	
 	<property name="def_torso_bottom_ref" value="0.0"/>
 	<property name="def_torso_side_ref" value="0.0"/>
+
+	<!-- laser front/rear  mount positions | relative to base_link -->
+	<property name="def_laser_front_x" value="0.32501"/>
+	<property name="def_laser_rear_x" value="-0.32501"/>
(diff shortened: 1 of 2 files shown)


fmessmer:
I see the need for this, however, I have some concerns about the way it is done...
The main problem is the way our base.xacro's are provided in our description packages, i.e. that the scanners are (hardcoded) parts of the bases without a chance of providing calibration/mounting parameters to them.

In particular, [this line](https://github.com/ipa320/cob_common/pull/156/files#diff-4a78351e68e5a32676bd919f1d3ecde4R7) somewhat introduces a dependency from cob_common to cob_robots....which is undesirable!

As we had in mind to restructure the base.xacro's anyway in the near future (e.g. provide separate xacros for the IPA Fahrdrehmodul instead of having the ancient PR2-wheel-based xacro), the calibration aspect for the laser scanner should be considered then, too.

In addition, as a minor comment, I would prefer to have full 6D calibration properties, i.e. x,y,z,roll,pitch,yaw rather than only x,y,yaw - even if z,roll and pitch are 0.0 for now...

fmessmer:
I created an issue to record this: https://github.com/ipa320/cob_common/issues/157

fmessmer:
do not merge - superseded by #305 

@ipa-frm

floweisshardt:
integrated into https://github.com/ipa320/cob_robots/pull/306

PR #300, opened 2015-05-11
added rplidar sensor to raw3-3 urdf and bringup

fmirus:


cob_bringup/robots/raw3-3.xml
```diff
@@ -63,6 +63,9 @@
 		<include file="$(find cob_bringup)/tools/base_collision_observer.launch" >
 			<arg name="robot" value="$(arg robot)" />
 		</include>
+
+		<include file="$(find rplidar_ros)/launch/rplidar.launch" />
```
fmessmer: Is `rplidar_ros` a publically available ROS-package?
If so, you can add a run_dependency to the package.xml.

fmirus: @ipa-fxm yes, the package can be found here: https://github.com/robopeak/rplidar_ros
I have also added the run dependency as suggested.

fmessmer: Thx, but `rplidar_ros` is not released, right?

mgruhler: Has been released into hydro, but not into indigo

floweisshardt:
if this is not released, which means "apt-get installable", we should be extremly carefull with adding this to cob_robots launch files as it will fail launching for normal users (without overlays).

My suggestion if you like to have this in:
- add a launch files argument , e.g. "include_non_released_packages" that will default to false
- put the rplidar.launch into a group which using "if" and "unless" features

mgruhler:
@ipa-frm Done?
@ipa-fmw FYI

fmirus:
@ipa-mig think so :wink:

mgruhler:
:ok:

cob_bringup/robots/raw3-3.xml
```diff
@@ -3,6 +3,7 @@
 
 	<!-- args -->
 	<arg name="robot" value="raw3-3"/>
+	<arg name="include_non_released_packages" default="false"/>
```
fmessmer: Can we use a more related arg name, e.g. `<arg name="use_rplidar" default="false"/>`?

cob_bringup/robots/raw3-3.xml
```diff
@@ -62,7 +63,12 @@
 		</include>
 		<include file="$(find cob_bringup)/tools/base_collision_observer.launch" >
 			<arg name="robot" value="$(arg robot)" />
-		</include>
+    </include>
+
+    <group if="$(arg include_non_released_packages)">
+      <include file="$(find rplidar_ros)/launch/rplidar.launch" />
+    </group>
+
```
fmessmer, line 71: tabs vs. spaces!

fmessmer:
I'll give this one a "good-to-go" if my two comments are considered....:wink:

cob_bringup/robots/raw3-3.xml
mgruhler: This has been done according to @ipa-fmw's comment.
So, what should we do? The current would be a more general approach, the other a more clear one.
I don't mind ;-)

floweisshardt: +1 for version from @ipa-fxm

mgruhler:
@ipa-frm push

fmirus:
@ipa-mig update

fmessmer:
LGTM

floweisshardt:
LGTM

PR #301, opened 2015-05-27
Fix raw3 3 launch and bringup package file

frederikhegger:


fmessmer:
Thanks, @frederikhegger 

In `indigo_dev` this fix is not required, as `cob_lookat_controller` is no longer used within `cob_bringup` and the missing launch file argument is already considered...

PR #315, opened 2015-07-13
missing run_dependency

dabarsch:
In debs, `cob_omni_drive_controller` is not installed...thus adding it as run_depend

### DIFF
--- a/cob_bringup/package.xml
+++ b/cob_bringup/package.xml
@@ -17,6 +17,7 @@
  
   <run_depend>cob_undercarriage_ctrl</run_depend>
   <run_depend>cob_base_drive_chain</run_depend>
+  <run_depend>cob_omni_drive_controller</run_depend>
   <run_depend>cob_lbr</run_depend>
   <run_depend>cob_head_axis</run_depend>
   <run_depend>cob_sound</run_depend>


fmessmer:
We should review the debs in cob_bringup and cob_bringup_sim...
...and **sort** them!!!
(maybe alphabetically and/or grouped according to which stack/repo they live in?)

fmessmer:
Seems like [this tool](https://github.com/DLu/roscompile) could help
See [here](http://lists.ros.org/pipermail/ros-users/2015-July/069537.html) for announcment and discussion

@ipa-fmw, @ipa-nhg, @ipa-mdl, @ipa-mig FYI

mathias-luedtke:
This tool cannot help here, since the dependency is hidden in a configuration file.
In addition the "type" param is quite common and we would need a very specialized tool that checks where the param is loaded to.
We should stick to system tests and use the released versions on all robots.

I would suggest to sort dependencies alphabetically, perhaps grouped by the XML tags, but not by stacks/repos. We move packages occasionally..

fmessmer:
Ok, well for me this PR was already a reminder to carefully review our `package.xml`'s in general.
I will test this tool and see whether it can help detect missing or obsolete dependencies...
Of course it doesn't help with the `type` dependencies...we'll need to take care of that manually.
Also, sorting alphatecially is fine for me, then....

fmessmer:
Will be resolved in #319...closing!

PR #326, opened 2015-08-18
making 'sim_enabled' a launch argument

### LINKED ISSUE #236: [cob_light] Handling `sim_enabled` param in simulation
See my comments about "Warning B" in https://github.com/ipa320/cob_simulation/issues/87#issuecomment-132086878

So far, we are using the same yaml for simulation and real hardware. This leads to the following warning: 

> [ WARN] [1439833261.548182269]: Serial connection on /dev/ttyLed failed.

In that yaml `sim_enabled` is set to false...thus, in simulation you cannot actually use `sim_enabled`

Maybe we should make it a private parameter that can be set via launch file argument in https://github.com/ipa320/cob_robots/blob/indigo_dev/cob_bringup/drivers/light.launch?

@ipa-bnm what do you think?

fmessmer:
@ipa-bnm 

allows to start simulation without Warning

> Serial connection on %s failed

fixes https://github.com/ipa320/cob_driver/issues/236

related to https://github.com/ipa320/cob_simulation/issues/87
@xwhm

### DIFF
--- a/cob_bringup/drivers/light.launch
+++ b/cob_bringup/drivers/light.launch
@@ -4,10 +4,12 @@
 	<arg name="robot" default="$(optenv ROBOT !!NO_ROBOT_SET!!)"/>
 	<arg name="pkg_hardware_config" default="$(find cob_hardware_config)"/>
 	<arg name="component_name" default="light"/>
+	<arg name="sim_enabled" default="false"/>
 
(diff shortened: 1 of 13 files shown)


benmaidel:
yes, this makes sense. lgtm

NadiaHG:
LGTM

PR #330, opened 2015-09-17
New teleop node

NadiaHG:


NadiaHG:
Has to be merged together with: https://github.com/ipa320/cob_command_tools/pull/109

cob_bringup/tools/teleop.launch
```diff
@@ -3,12 +3,15 @@
...
 	<arg name="pkg_hardware_config" default="$(find cob_hardware_config)"/>
-	<arg name="version" default="v2"/>
 
-	<include file="$(find cob_bringup)/tools/teleop_$(arg version).xml">
+	<include file="$(find cob_bringup)/drivers/joy.launch">
 		<arg name="robot" value="$(arg robot)" />
 		<arg name="pkg_hardware_config" value="$(arg pkg_hardware_config)" />
 	</include>
 
+	<node pkg="cob_teleop" type="cob_teleop" name="$(anon teleop_node)" ns="teleop" output="screen">
```
fmessmer: order: `ns` `pkg` `type` `name`!

cob_bringup/tools/teleop.launch
```diff
@@ -3,12 +3,15 @@
...
+		<rosparam command="load" ns="/teleop" file="$(arg pkg_hardware_config)/$(arg robot)/config/teleop.yaml"/>
```
fmessmer: What kind of parameter should this be?
private? as you put it within the `node`-tag?
or global? as you use `ns=/teleop`?
I guess the roscpp calls `getParam()` in the node need to be used with the correct NodeHandle to resolve this!

I'd prefer to remove `ns=/teleop` in the `rosparam` tag and therefore use a private NodeHanlde nh_priv("~") in the cpp node!

cob_hardware_config/cob3-2/config/teleop.yaml
```diff
@@ -1,49 +1,62 @@
...
-# axes
-axis_vx: 1
-axis_vy: 0
-axis_vth: 2
-up_down: 5 #tray--up/down; tilt--front/back, here we just name up_down
-left_right: 4 #pan--left/right
+#mode2: Trajectory controller
+torso_sss_default_target_button: 1
+head_sss_default_target_button: 3
+arm_sss_default_target_button: 2
```
fmessmer, line 30: sort this by number?
(apply this for other yaml files, too!)

cob_hardware_config/cob3-2/config/teleop.yaml
```diff
@@ -1,49 +1,62 @@
...
-upper_neck_button: 4
-tray_button: 3
-arm_joint12_button: 0
-arm_joint34_button: 1
-arm_joint56_button: 2
-arm_joint7_button: 3
-base_safety_button: 4
 deadman_button: 5
+safety_button: 4
+init_button: 9
```
fmessmer, line 22: sort this by number?
(apply this for other yaml files, too!)

cob_hardware_config/cob3-2/config/teleop.yaml
```diff
@@ -1,49 +1,62 @@
...
+    twist_topic_name: '/base_controller/command',
+    twist_max_velocity: [0.3, 0.2, 0.3],
+    twist_max_acc: [0.5, 0.5, 0.7]
   },
   torso: {
-    joint_names: ['torso_lower_neck_pan_joint','torso_lower_neck_tilt_joint','torso_upper_neck_pan_joint','torso_upper_neck_tilt_joint'],
-    joint_step: [0.05,0.05,0.075,0.075]
+    sss_default_target: "home",
+    velocity_topic_name: "/torso/joint_group_velocity_controller/command",
+    joint_velocity: [0.05,0.05,0.05,0.05]
```
fmessmer: use a space for clearer separation
(apply this for other yaml files, too!)

cob_hardware_config/cob3-2/config/teleop.yaml
```diff
@@ -1,49 +1,62 @@
...
   arm: {
-    joint_names: ['arm_1_joint','arm_2_joint','arm_3_joint','arm_4_joint','arm_5_joint','arm_6_joint','arm_7_joint'],
-    joint_step: [0.1,0.1,0.1,0.1,0.1,0.1,0.1]
-  },
-  gripper: {
-    joint_names: ['sdh_finger_11_joint', 'sdh_finger_12_joint', 'sdh_finger_13_joint', 'sdh_finger_21_joint', 'sdh_finger_22_joint', 'sdh_finger_23_joint', 'sdh_thumb_2_joint', 'sdh_thumb_3_joint'],
-    joint_step: [0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0]
+    sss_default_target: "home",
+    velocity_topic_name: "/arm/joint_group_velocity_controller/command",
+    joint_velocity: [0.05,0.05,0.05,0.05,0.05,0.05,0.05]
```
fmessmer: use a space for clearer separation
(apply this for other yaml files, too!)

cob_hardware_config/cob3-2/config/teleop.yaml
```diff
@@ -1,49 +1,62 @@
...
+#mode3: Velocity group controller
+arm_joint1_button: 0
+arm_joint2_button: 1
+arm_joint3_button: 2
+arm_joint4_button: 3
 
-# component configuration
-modules: {
+up_down_button: 5
+right_left_button: 4
```
fmessmer, line 39: sort this by number?
(apply this for other yaml files, too!)

cob_hardware_config/cob3-2/config/teleop.yaml
NadiaHG, line 22: why? the number is the parameter, not the parameter name and it depends on the joystick, if we sort it by number we will have a different parameters order for each robot. 
And for me it is better define always first the deadman_button , without this parameter you can't move any component, and then the other buttons that are not required.

cob_hardware_config/cob3-6/config/teleop.yaml
```diff
@@ -1,49 +1,62 @@
...
+#mode2: Trajectory controller
+torso_sss_default_target_button: 1
+head_sss_default_target_button: 3
+arm_sss_default_target_button: 2
+
+#mode3: Velocity group controller
+arm_joint1_button: 0
+arm_joint2_button: 1
+arm_joint3_button: 2
+arm_joint4_button: 3
```
fmessmer, line 36: Is it only possible to command 4 of 7 joints with the new teleop?

Could we use a bitmap-like approach to increase the number of joints? such as [0|1]_2^0 + [0|1]_2^1 + [0|1]_2^2 + [0|1]_2^3. This would give us the chance to command 15 joints

However, if we can only command maximum four joints **and** the bitmap-like approach is not really intuitive to use, I'd rather remove this mode (i.e. single-joint-motion) from the teleop...

cob_hardware_config/cob4-1/config/teleop.yaml
```diff
@@ -1,20 +1,97 @@
...
-base_safety_button: 4
 
-# axes
-axis_vx: 1
-axis_vy: 0
-axis_vth: 2
+#mode2: Trajectory controller
+torso_sss_default_target_button: 1
+sensorring_sss_default_target_button: 3
+head_sss_default_target_button: 3
```
fmessmer, line 30: is it ok to use the same button for two components? both head and sensorring on `3`?

cob_hardware_config/cob4-1/config/teleop.yaml
```diff
@@ -1,20 +1,97 @@
...
+sensorring_joint1_button: 1
+head_joint1_button: 2
+arm_right_joint1_button: 0
+arm_right_joint2_button: 1
+arm_right_joint3_button: 2
+arm_right_joint4_button: 3
+arm_left_joint1_button: 0
+arm_left_joint2_button: 1
+arm_left_joint3_button: 2
+arm_left_joint4_button: 3
```
fmessmer, line 45: can't we reuse this parameter?
such as only call it `joint1_button` and then get the name via the `joint_names` parameter using the according index, i.e. joint1_button ==> [0]?

NadiaHG, line 45: No, because it depends on your robot and your joystick configuration. It has not be joint1_button always 0, for example for the head is the joint1_button is 2.

cob_hardware_config/cob4-1/config/teleop.yaml
NadiaHG, line 30: Yes, it is right

floweisshardt:
visual review: OK

cob_hardware_config/cob3-2/config/teleop.yaml
fmessmer, line 22: fine with me then

PR #336, opened 2015-09-28
corrected light_configuration.yaml

benmaidel:


floweisshardt:
LGTM

fmessmer:
Could you please apply this fix to the other robots, too? Thanks!
- cob3-2
- cob3-6
- cob3-9
- cob4-2

benmaidel:
this do only apply to raw3-3, so no need to change it for other robots

fmessmer:
Ok, thanks...merging...

PR #338, opened 2015-09-28
start cob_script_server at bringup because new teleop node needs it

benmaidel:


fmessmer:
Could you apply this for the other robots as well?
In order to avoid lots of copying...should we add the launch file e.g. to the [upload_param.launch](https://github.com/ipa320/cob_robots/blob/indigo_dev/cob_default_robot_config/upload_param.launch)? Because without those parameters, the script server could not do anything...and vice-versa....

benmaidel:
already applied to all raw robots (look at files changed).
I think it is not the best idea to start the script server node inside a launch file called upload_param.launch. With the new cob_teleop node, it is not possible to teleop, init, recover etc. the robot with the joypad without running the script server. So the script server became an essential part of the bringup procedure. Perhaps upload_param.launch and script_server.launch should be placed inside robot.launch ?

fmessmer:
I saw you added it to all raw's...question is: should this be added to all cob's, too? @ipa-fmw?
You got a point with your argumentation about script_server now being an essential part of bringing up a robot...so fine with me adding it to the robot xml's?

floweisshardt:
it is already part of cob4-2.xml see https://github.com/ipa320/cob_robots/blob/indigo_dev/cob_bringup/robots/cob4-2.xml#L87. no objection from my side to include it into all robot xmls files.
we shouldn't place it into upload_param.launch, let's keep that separated.

floweisshardt:
LGTM, can be merged

benmaidel:
but whats about the cob3's? Teleop will not work without it

floweisshardt:
right. Can you please add it to all cob3s too?

benmaidel:
can be merged now

floweisshardt:
thanks

PR #339, opened 2015-09-28
do not respawn phidgets driver

benmaidel:
because if no phidget is connected the driver will restart all the time


fmessmer:
LGTM

floweisshardt:
LGTM

PR #343, opened 2015-10-25
Update cob3-9

NadiaHG:
-Updated the config files for cob3-9 to the new drivers version
-Calibrated the arm , using the right joint orientation convention (also changed using the Schunk MTS software)


NadiaHG:
test on cob3-9: OK

NadiaHG:
Has to be merged with : https://github.com/ipa320/schunk_modular_robotics/pull/148

cob_default_robot_config/cob3-9/command_gui_buttons.yaml
```diff
@@ -46,15 +46,6 @@ group5:
     [home,move,home],
     [folded,move,folded],
     [pregrasp,move,pregrasp]]
-group6:
```
floweisshardt, line 49: why is this section removed from dashboard? It is necessary on all cob3s to move the arm to the frontside of the robot.

floweisshardt:
LGTM, except for the removed arm trajectories, see comment above

cob_default_robot_config/cob3-9/command_gui_buttons.yaml
NadiaHG, line 49: The frontside positions are not defined , look https://github.com/ipa320/cob_robots/blob/indigo_dev/cob_default_robot_config/cob3-9/arm_joint_configurations.yaml and , if you take a look of the history, these positions were never defined. When we delivered cob3-8 and cob3-9 robots, we decided together remove these positions to the default setup of the cob3s.

fmessmer:
LGTM

PR #337, opened 2015-09-28
raws bases still use `/base_controller` namespace instead of `/base/driver`

benmaidel:

PR #331, opened 2015-09-22
[WIP] Allow multi-robot simulation

ipa-srd-fg:
- delete leading slashes -> relative path 
- renaming laser scan topics
- use `robot_id` in gazebo plugin

PR #346, opened 2015-11-02
[WIP] Towards Multi-Robot-Support (NOT Multi-Robot-Support)

fmessmer:
Supersedes #331 and #337 
Considers all remarks from #331
Includes additional changes to be compatible to https://github.com/ipa320/cob_common/pull/174

I'm going to test this with the real hardware tomorrow...


fmessmer:
ToDo:
- ~~update .rviz configs according to new topic names~~ :point_right: done

Call for feedback:
Simulated diagnostics are all wrong for most of the robots...they do not match the patterns defined in the aggregator yaml....I really tempted to rather delete diagnostics in simulation than fix them:
- raw's do not use simulated diagnostics at all
- cob's simulated diagnostics have only been introduced to "make the light green" in the dashboard
- they are consuming quite some CPU, and take almost forever to spawn
- they do not provide any meaningful data
  Comments?

:point_right: We'll remove it in a separate PR after this has been merged!

fmessmer:
Some insights/problems from testing on cob4-2:
- ~~/safety_controller/base_controller/odometry is not subscribing to /base/odometry_controller/odometry~~
- ~~`/safety_controller/diagnostics` and `/cob_sound/diagnostics` are not properly remapped~~
- ~~service_ns `base/driver` resolves to `/command_gui/torso/driver/init`~~
- containt: "base" in diagnostics is not sufficient, as `base_laser_X` contains "base", too

cob_bringup/robots/cob3-2.xml
```diff
@@ -42,15 +42,15 @@
 		<!-- lower robot -->
 		<include file="$(find cob_bringup)/drivers/sick_s300.launch" >
 			<arg name="robot" value="$(arg robot)" />
-			<arg name="name" value="front" />
+			<arg name="name" value="base_laser_front" />
```
benmaidel, line 45: what do you think about namespacing like we have done it with the controllers. Eg: `base/laser/front`

fmessmer:
I had to undo some of the "global / removal" changes introduced by @ipa-srd-fg as it did not work well for  cob4-2 without global `/`, e.g.
- cob_sound and safety_controller do not publish to global but to local diagnostics topic
- topic_ns and service_ns in cob_default_robot_config gets resolved to /command_gui/... without global /

Still, I think this (set of) PR introduces quite a few good contributions/cleanups/simplifications....I was able to "operate" most of cob4-2's components, but more testing would be required!
@ipa-nhg could you take care of testing it more sorrowfully on cob4-2 and maybe a cob3 and a cob4 simulation?
@ipa-srd could you take care of testing it with a raw3?

For proper testing, you would need latest indigo_dev branches for most of the packages, as well as my `feature/multi_robot` branches from the following repos:
- cob_common
- cob_driver
- cob_robots

Once the bringup layer is stable....navigation repos should be adjusted accordingly....

I'll be out-of-the-office until 16th...but can reply to github discussion!

cob_bringup/robots/cob3-2.xml
fmessmer, line 45: I chose this name as it is analog to the link name which is base_laser_X_link....also, I think a none-slash'ed **name** is more convenient than a slash'ed name...

NadiaHG:
This pull request mixes different features, and also it will not work for a multirobot environment. Is it not better if we open a pull request pro feature? I though that we decided try to make the pull requests as concrete as possible. 

We can divide this pull request into:
- upload the base namespaces /base_controller.. to /base/driver... : this pull request is already open https://github.com/ipa320/cob_robots/pull/337/files and tested on the robot , we can merge it.
- Renamed the laser scanner and the camera topics and links, still not tested but it looks good for me , this pull request will require https://github.com/ipa320/cob_common/pull/174 and the navigation and perception packages have to be adapted. 
- The fake diagnostics parameters , it will only affect the simulation
- Updated cartesian controller parameter (for example https://github.com/ipa320/cob_robots/pull/346/files#diff-b0b2f206f0d7c50ca51f6535641fc322)
- removed global namespace : from my point of view this feature is at the moment not mergeable , it doesn't work for a multirobot environment. Also this feature has to be carefully tested, maybe it will require some driver modifications... Some launch files(drives) require global namespaces , it is not removable for all the cases , in order to make it working at the moment on the hardware we will have a mix between global namespaces and local namespaces.

cob_bringup/robots/cob3-2.xml
mathias-luedtke, line 45: But at the topics should be moved into the base namespace..

mathias-luedtke:
+1 for splitting

benmaidel:
+1 for splitting this PR

+1 consistent topic namespaces `base/laser/...`

+1 for merging PR #337 first (this one is tested and open for over a month, do not superseed it with this big PR)

fmessmer:
I moved some less disputed parts of this PR into PRs of their own (#351, #352, #353, #354, #355, #356)

The rest of this PR is related to global namespaces, laser topic names, etc. and will be discussed according to the architecture discussion....

floweisshardt:
already merged your PRs

fmessmer:
updated...

mathias-luedtke:
I just scrolled down the change list (still too many files for a review). The leading slash must not be removed from remaps if the node is running in a namespace. The namespace might even be introduced in the including file.

cob_bringup/drivers/hokuyo.launch
```diff
@@ -9,9 +9,7 @@
 	<rosparam ns="laser_$(arg name)" command="load" file="$(arg pkg_hardware_config)/$(arg robot)/config/laser_$(arg name).yaml"/>
 
 	<!-- start laser driver -->
-	<node ns="laser_$(arg name)" pkg="hokuyo_node" type="hokuyo_node" name="laser_$(arg name)" respawn="false" output="screen">
```
ipa-mdl: This a non-trivial change and does not belong to a multi-robot feature.

cob_bringup/drivers/hokuyo.launch
```diff
@@ -9,9 +9,7 @@
...
-		<remap from="scan" to="/scan_$(arg name)_raw"/>
-	</node>
+	<node ns="laser_$(arg name)" pkg="hokuyo_node" type="hokuyo_node" name="laser_$(arg name)" respawn="false" output="screen"/>
```
ipa-mdl: This a non-trivial change and does not belong to a multi-robot feature.

cob_bringup/drivers/hokuyo.launch
```diff
@@ -9,9 +9,7 @@
...
```
mathias-luedtke, line 12: This a non-trivial change and does not belong to a multi-robot feature.

fmessmer:
Again:
**This PR does not provide Multi-Robot-Support!**
It simply tries to play back a couple of insights gained during the experiments performed by @ipa-srd-fg **towards** Multi-Robot-Support!

Mainly the many hard-coded global namespaces have been blocking the usage of the current (Single-Robot) setup in a scenario with several robot instances! In case a current hardcoded topic is needed in a remap, this will then probably lead to problems again and we might need to find a solution with which topics do not need to be remapped globally...

However, putting each robot in a new namespace of its own did not work well enough due to other problems faced, mainly regarding tf_prefix and more hard-coded global namespaces in non-ipa320-related code. This is why it has been decided to hold this approach and try the Multi-Master-Approach (currently investigated by @ipa-srd-fg).

However and still, I was thinking that _some_ of the changes of #331 would be helpful for our standard Single-Robot setup, too, which is why I created this PR here

Due to the discussion about Architecture and not yet agreed consistent topic names and such, **this PR is not meant to be merged** (yet)....
I only was merging with latest ipa320/indigo_dev to resolve the merge conflicts and to reduce this PR as my single feature PRs #351 - #356 have been merged!
I am willing to further reduce this step by step as we agree on how to deal with this...and finally this PR will be squashed or even closed anyway....until then this PR can be seen as a reminder or simply for discussion.

---

And "too many files" does not necessarily hold for cob_robots! 
Changing something for one robot sometimes needs to be changed in several files for even the same robot....and we have **13 robots** both in simulation and hardware!

cob_bringup/drivers/hokuyo.launch
fmessmer, line 12: As soon as we agree on what topic name to use for laser scanners, this feature will be proposed in a separate PR and reduced from this PR...

mathias-luedtke:
> this PR is not meant to be merged

Then we should mark it or lock it.

> (yet)

It should not be merged (as it is now) because it cannot work.

> And "too many files" does not necessarily hold for cob_robots! 
> Changing something for one robot sometimes needs to be changed in several files for even the same robot....and we have 13 robots both in simulation and hardware!

Agreed. "too many files" was not what I actually meant.
Since we have so many robots the commits should include one feature for all robots and not different features for one robot.

So for example "stripped leading slashes from frame names" would be nice for review.
The reviewer could review this single commit only and check if all frame are touched an nothing else.
This way we can review hundreds of files and thousands of changes.

fmessmer:
Restructuring the laser scanner topics is done in a fresh PR #376 
Everything else changed within this PR here, has been discussed and agreed to be discarded i.e. global namespaces is kept for now (for both topics and link names).

The only things left would be the restructuring of the camera sensor topics. However, I will provide a PR once  https://github.com/ipa320/cob_common/pull/179 has been finalized...

Thus, I will close this PR (finally)

PR #347, opened 2015-11-03
removed leading / from tf frame names. 

benmaidel:
They are no longer supported in tf2.
If you use a tf2 you will get the following error if you use leading slash in tf name:
`Transform failure: Invalid argument "/torso_cam3d_left_link" passed to lookupTransform argument source_frame in tf2 frame_ids cannot start with a '/' like`

### DIFF
--- a/cob_bringup/drivers/openni.launch
+++ b/cob_bringup/drivers/openni.launch
@@ -13,8 +13,8 @@
 	<include file="$(find openni_launch)/launch/openni.launch">
 		<arg name="camera" value="$(arg name)"/>
 		<arg name="device_id" value="$(arg device_id)" />
-		<arg name="rgb_frame_id" value="/$(arg name)_link" />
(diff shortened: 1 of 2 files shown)


fmessmer:
I guess beside that issue, the cam3d topic names need a significant review anyway as topics in simulation do not match the topic names on real hardware for most of the kinect/softkinetic/prosilica sensors....  
As #346 tries to introduce a consistent naming in simulation...can we tackle this issue after #346 has been merged and then try to remove trailing slashes for link_names wherever it applies?

benmaidel:
I think you mixed two things together. This PR has nothing to do with cam3d topic names and it is also not related to #346. This is only a fix for the openni launch. This PR is fine to merge without considering any other PRs.

floweisshardt:
I'll merge this and open a new issue for reviewing camera topic names.

floweisshardt:
new issue is https://github.com/ipa320/cob_robots/issues/348

PR #355, opened 2015-11-08
fix and consistent services and topics in base config

fmessmer:
fix names for cob3's and make config files consistent
(split this feature from #346)


NadiaHG:
LGTM

floweisshardt:
LGTM

PR #356, opened 2015-11-08
remove lookat

fmessmer:
as it is not used/supported anymore
(split this feature from #346)


mathias-luedtke:
Is it superseded by the new twist controller?
Looks like other relics are spread all over ipa320..
At least lookat.urdf.xacro should be deleted or do you really want to keep it?

fmessmer:
cob_twist_controller does (**not yet**) provide the lookat functionality...but when it does, it will not require anything to be added to the urdf (hopefully)...
However, it hasn't been working anymore based on the URDF (whose leftover-changes have been removed in this PR)!

@ipa-mdl: what other occurrences did you spot?
For testing, I would like to keep the lookat.urdf.xacro in cob_common for the time being...

floweisshardt:
LGTM

PR #358, opened 2015-11-10
remove simulated fake_diagnostics

fmessmer:
this removes all the simulated fake_diagnostics...
...as they were not compatible with their respective real robot hardware configuration
...as raw's did not use them so far

Diagnostics in the dashboard now just are STALE
sim_dashboard_stale


floweisshardt:
LGTM

NadiaHG:
LGTM

PR #359, opened 2015-11-10
[WIP] add sensorring sensors within cob_hardware_config

fmessmer:
This is related to [this discussion](https://github.com/ipa320/cob_common/pull/175#issuecomment-154128294) from https://github.com/ipa320/cob_common/pull/175

By adding the sensors like this, the original sensorring.urdf.xacro in cob_common (i.e. without sensor) can be used without any changes in cob_common (i.e. the sensorring_XXX.urdf.xacro could be removed again). In case another/additional sensor is mounted to the sensorring (or anywhere else on the robot), only the new file `cob4-2.sensors.xacro` needs to be modified (and of course according launch files in cob4-2.xml).

I'm open for discussion about how to name this file....maybe it could make sense to name it (non-sensor-related) e.g. cob4-2.[additons|aux|...].xacro. If we name it in a more general way, it might not necessarily be limited to sensors...but could be used for other components, too...
Doing it like that could also be applied to the other robots...i.e. raw3-1 has a lot of "attachements" added within raw3-1.urdf.xacro....by moving those into the additional file...the main file would look "cleaner"...

What do you think?
Agreement could be discussed on Monday!

### DIFF
--- a/cob_hardware_config/cob4-2/urdf/cob4-2.sensors.xacro
+++ b/cob_hardware_config/cob4-2/urdf/cob4-2.sensors.xacro
@@ -1,0 +1,11 @@
+<?xml version="1.0"?>
+<robot xmlns:xacro="http://www.ros.org/wiki/xacro">
+
+  <!-- sensorring sensors -->
+  <xacro:include filename="$(find cob_description)/urdf/sensors/kinect.urdf.xacro" />
+
+  <xacro:cob_kinect_v0 name="sensorring_cam3d" ros_topic="sensorring_cam3d" parent="sensorring_link">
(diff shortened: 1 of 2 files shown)


NadiaHG:
I would like to define the whole robot together, for me a camera is a component , like an actuator, 

My suggestion was define a new level between robot and component: modules. And also define a default configuration for each module.  The robots bringup should use (as links) the default configuration of each module. Because also ,at the moment, I have the problem that if I create a new robot I have to COPY all the configuration files , and if we modify a feature we have to adapt the configuration files for all the robots.

I tried to create a tree structure of my idea:

screenshot from 2015-11-13 11 44 36

fmessmer:
Let's keep this locked and unmerged until we discussed how to simplify the configuration structure during one of our next orga sessions....

fmessmer:
closing in favor of issue #437

PR #361, opened 2015-11-12
fixed/cleaned raw3-3 config

benmaidel:
removed old unnecessary frida stuff

### DIFF
--- a/cob_hardware_config/raw3-3/config/robot_modules.yaml
+++ b/cob_hardware_config/raw3-3/config/robot_modules.yaml
@@ -1,1 +1,1 @@
-robot_modules: ['base','arm']
+robot_modules: ['base','torso']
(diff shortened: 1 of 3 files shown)


benmaidel:
tested on raw3-3!

floweisshardt:
LGTM

NadiaHG:
LGTM

cob_hardware_config/raw3-3/config/robot_modules.yaml
```diff
@@ -1 +1 @@
-robot_modules: ['base','arm']
+robot_modules: ['base','torso']
```
mgruhler: What is the torso on raw3-3? The linear actuators on top? Can this be moved via teleop (this is what `robot_modules.yaml` is for, right?) or why do we need to add this here?

benmaidel: In fact a mistake. Mixed raw3-1 and raw3-3 together.

fmessmer:
Formal Review OK

PR #363, opened 2015-11-12
added led offset param to torso light config

benmaidel:


floweisshardt:
LGTM

NadiaHG:
LGTM

cob_hardware_config/cob4-2/config/light_torso.yaml
```diff
@@ -2,6 +2,7 @@ devicestring: /dev/ttyLED
 devicedriver: stageprofi
 baudrate: 38400
 num_leds: 58
+led_offset: 5
```
fmessmer, line 5: Is there a default value `0`?
Do the light yamls for the other robots need this new parameter, too?

benmaidel, line 5: nope, they don't. Default is always 0.

PR #367, opened 2015-11-17
[hotfix] Scan unifier config files missed

NadiaHG:


fmessmer:
LGTM

---

On a side note:
@ipa-fmw thought that 100Hz might be faster than the actual scanner publish rate....Should the rate be reduced?
@ipa-bnm What do you think?

mgruhler:
CR: :ok:

@ipa-fxm @ipa-fmw @ipa-bnm  Yes, the rate of 100Hz is (considerably) faster than the scanner publish rate (which is 12Hz for the S300, afaik). However, this is on purpose to send out the combined scan asap.
It is not deterministic in which order the single scan topics come in and it is, in the worst case, almost one full scan rate cycle delay. This is due to the scan_unifier waiting until it received information from _every_ scan before it combines one.
This frequency was chosen to be high to check the subscriber callbacks frequently to introduce as little latency between the original scans and the combined one as possible.
However, it could still be beneficial to change this frequency...

fmessmer:
Moving discussion about the update rate to https://github.com/ipa320/cob_driver/issues/248

Pull request can be merged!

benmaidel:
CR: ok

floweisshardt:
FR: OK

PR #368, opened 2015-11-17
Parameter config updates

fmessmer:
updates for cartesian_control after ipa-fxm-mb's final presentation on cob4-2 hardware
consistently applied to all robots

### DIFF
--- a/cob_bringup/controllers/generic_cartesian_controller.launch
+++ b/cob_bringup/controllers/generic_cartesian_controller.launch
@@ -14,7 +14,10 @@
 		<rosparam command="load" file="$(arg pkg_hardware_config)/$(arg robot)/config/$(arg component_name)_self_collision.yaml"/>
 
 		<!-- Cartesian controllers -->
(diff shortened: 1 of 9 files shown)


floweisshardt:
FR: OK

NadiaHG:
FR: OK
Tested on the robot?

benmaidel:
FR: :ok:
can not tell anything about the functionality. Just one remark: The pid gains you use, seems to be randomly selected(use the half proportional value as before)?! Is this right?

fmessmer:
The parameter changes are related to feature updates in https://github.com/ipa320/cob_control/pull/63

The PID were originally tuned in simulation but then proofed to be too wild for real hardware...(tuning is still an ongoing process while more and more tests are being performed with the twist_controller on real hardware)

fmessmer:
ToDo:
- add twist_mux input topic for `cob_twist_controller`:`/base/twist_mux/command_syncmm`

PR #369, opened 2015-11-17
Feature cob4 1 without arms

fmessmer:
builds on top of #368 (respective commits will be removed after #368 is merged)

As discussed: 
For our new project, we use cob4-2 without arms. In order to simplify/separate configuration, we use cob4-1 in the required setup (cob4-1 was planned to be removed anyway).
Therefore, I meld'ed and merged the latest cob4-2 setup into cob4-1 setup and removed everything related to head, arms and gripper in a single commit (https://github.com/ipa320/cob_robots/commit/c1f2bf6cbb4403da798c37b11bed9baea6c6424d)

requires: https://github.com/ipa320/cob_calibration_data/pull/93

PR #370, opened 2015-11-17
Remove velocity controllers for elmo devices

fmessmer:
includes #368 and #369 

removes velocity controllers for elmo devices as velocity mode is dangerous because joint limits are not considered correctly!


cob_default_robot_config/cob4-1/command_gui_buttons.yaml
```diff
@@ -4,45 +4,13 @@
 #  buttons: [[button_name1,function_name1,parameter_name2,
 #    [button_name2,function_name2,parameter_name2]]
 
-group1:
+group10:
```
benmaidel, line 7: why do you change the groups this drastically? is there a system behind?

fmessmer, line 7: For cob4-2, there are more than 10 columns for knoeppkes.
According to @ipa-nhg, the ordering was faulty when using single-digit groups (0-9) together with dual-digit groups (10-99)...
The values have been copied from the current cob4-2 config...

benmaidel, line 7: than this seems to be a bug on android gui side and/or cob_command_tools side.

floweisshardt, line 7: in the command_gui the list is just sorted alphabetically, which means the following order

group1
group11
group1000000043
group2
group3

Switching to two-digit groups is OK for me

PR #375, opened 2015-11-20
modified travis config to build with source overlays

### LINKED ISSUE #374: Add launch file checks and URDF checks
We need to write these tests and add them to travis

@ipa-fmw @ipa-mdl 
Do you have a good idea how to deal with the problem about using `$(find )` and `$(arg )` and such mixed together?

mathias-luedtke:
closes #374

### DIFF
--- a/.travis.rosinstall
+++ b/.travis.rosinstall
@@ -1,0 +1,57 @@
+- git:
+    uri: 'https://github.com/ipa320/cob_calibration_data.git'
+    local-name: cob_calibration_data
+    version: indigo_dev
+- git:
+    uri: 'https://github.com/ipa320/cob_command_tools.git'
+    local-name: cob_command_tools
+    version: indigo_dev
+- git:
+    uri: 'https://github.com/ipa320/cob_common.git'
+    local-name: cob_common
+    version: indigo_dev
+- git:
+    uri: 'https://github.com/ipa320/cob_control.git'
+    local-name: cob_control
+    version: indigo_dev
+- git:
+    uri: 'https://github.com/ipa320/cob_driver.git'
+    local-name: cob_driver
+    version: indigo_dev
+- git:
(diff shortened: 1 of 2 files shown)


floweisshardt:
looks good on travis. As soon as we have reasonable tests defined, we should unkomment 'catkin_test_results' to reflect the test result status in the travis log. @ipa-mdl: Thanks!

mathias-luedtke:
There is an issue with the latest `catkin` version that results in many cmake errors/warnings (https://github.com/ros/catkin/issues/761).

mathias-luedtke:
Functionality was successfully tested in #378.

@ipa-fxm, @ipa-fmw: your turn :D

fmessmer:
Awesome!
The run_test failures detected in #378 are fixed in https://github.com/ipa320/cob_robots/pull/379

I assume, we now should apply the same `.travis.yml` and `.travis.rosinstall` to all the other repos, too?
(Note that in the `.travis.rosinstall`, entries for all bringup repos should be included **except for the respective repo itself** - as explained by @ipa-mdl)

mathias-luedtke:
> Note that in the .travis.rosinstall, entries for all bringup repos should be included except for the respective repo itself

The respective repo can be included, but it will be build twice then.
If you don't bundle a `.travis.rosinstall` file the default one is used (https://github.com/ipa320/setup/blob/master/repositories/cob_indigo_bringup.rosinstall).

fmessmer:
Of course....travis.rosinstall is installed in an underlay...
We should definitely update the default .rosinstall file in setup according to the changes in https://github.com/ipa-fxm/cob_robots/commit/edf9f96a78168ab8c7e3cf1fd18f1da205e4f21d (then we might be able to just use that one)

floweisshardt:
CR: OK

PR #376, opened 2015-11-20
restructure laser topics

fmessmer:
This PR goes along with https://github.com/ipa320/cob_common/pull/178

It changes the laser scanner topics according to the convention as discussed today, i.e.
`/base_laser_X/scan_raw` (raw topic from hw-driver and simulation), `/base_laser_X/scan_filtered` (scan filtered)

I hope I did not forget anything...of course in case the two PRs are merged, also `cob_navigation` and co. need to be adjusted (but I think this is already prepared)....


fmessmer:
Note: This PR is based on `indigo_dev` (as of Friday, 20th). In case other cob_robots PRs get merged, this PR might need an update (e.g cob4-3!)

mgruhler:
Hmm, maybe I missed that decision on Friday.
But didn't we decide to go for `/base_laser_X/scan` instead of `/base_laser_X/scan_filtered`?

This is actually a minor issue, but the "filter" that we use here, is more like a "cropper". It does not do any "real filtering" (whatever that might be ;-) ).

The reason I am asking is that there actually are real filters (e.g. the [laser_filters](http://wiki.ros.org/laser_filters) package) which can do intensity filtering, mean/median filters and shadow filtering.
Those work actually quite well and it might be worth investigating those further.

How should this be called then?

mgruhler:
oh, and `cob_navigation` is not yet prepared ;-)

mgruhler:
and you missed the `scan_unifier_config.yaml`s

floweisshardt:
+1 for `/base_laser_X/scan` (instead of `_filtered`). See architecture document in https://github.com/ipa320/orga/pull/1

fmessmer:
- renamed `/base_laser_X/scan_filtered` to `/base_laser_X/scan` as in https://github.com/ipa320/orga/pull/1
- cob_navigation is prepared in https://github.com/ipa320/cob_navigation/pull/59
- cob_scan_unifier yamls will be updated in a new PR after the merge session

fmessmer:
wrt https://github.com/ipa320/cob_robots/pull/376#issuecomment-158867469

Now I remember why I chose `scan_filtered` in the first place: It was the default output of the laser_scan_filter node. In order to avoid total confusion, we now use the convention as decided in https://github.com/ipa320/orga/pull/1...minor adjustments have been applied in https://github.com/ipa320/cob_driver/pull/251

fmessmer:
closing in favor of https://github.com/ipa320/cob_robots/pull/383

PR #381, opened 2015-11-23
Use 10 threads for wstool init

mathias-luedtke:
Bad timing.


floweisshardt:
FR: OK

mathias-luedtke:
Saves 10 seconds.
Feel free to merge it, but it is not necessary.

PR #377, opened 2015-11-20
introduced twist_mux and new velocity smoother

benmaidel:
- added twist_mux launch files and configs as symlinks (common folder)
- added new velocity_smoother launch files and configs
- adjusted topic names to new structure

coupled with:
- https://github.com/ipa320/ipa_navigation/pull/87
- https://github.com/ipa320/cob_navigation/pull/57
- https://github.com/ipa320/cob_command_tools/pull/125
- https://github.com/ipa320/cob_control/pull/67

successfully tested on raw3-3

PR #382, opened 2015-11-24
Feature/twist mux vel smoother

floweisshardt:
minor fixes to #377  to adapt to https://github.com/ipa320/orga/pull/1#issuecomment-159195427


cob_hardware_config/cob4-2/config/teleop.yaml
```diff
@@ -2,6 +2,7 @@
 ##########
 # common params
 run_factor: 2.0
+apply_ramp: false
```
NadiaHG, line 5: why is apply_ramp only defined for cob4-2 and raw3-3?

floweisshardt, line 5: good point. Adapted to all robots in https://github.com/ipa320/cob_robots/commit/a342522e2c05822f94cc843001bd6ae0f6c2380f

PR #365, opened 2015-11-16
added cob4-3

NadiaHG:

PR #385, opened 2015-11-24
added cob4-3 - updated

fmessmer:
includes #365 
and applies changes according to today's twist_mux, vel_smoother and laser scanner topic activity


floweisshardt:
CR: OK

PR #390, opened 2015-12-02
[hotfix] remove head and arms from teleop config

floweisshardt:
cob4-1 currently has no arms and no head. This hotfix is needed to init and recover by joystick.


NadiaHG:
CR:  OK

floweisshardt:
Merging monday...

PR #394, opened 2015-12-11
remove env config in all robot launch files

### LINKED ISSUE #373: [cob_bringup] remove ROBOT_ENV/robot_env dependency
See discussion started [here](https://github.com/ipa320/cob_robots/pull/369#discussion_r45443892)

floweisshardt:
we do not want a dependency to env config for the robot launch files, thus removing.
Navigation goals need to be uploaded separately (e.g. in scenario.launch, but not in robot bringup)

fixes #373

### DIFF
--- a/cob_bringup/robot.launch
+++ b/cob_bringup/robot.launch
@@ -3,12 +3,7 @@
 
 	<param name="/use_sim_time" value="false"/>
 	<arg name="robot" default="$(optenv ROBOT !!NO_ROBOT_SET!!)"/>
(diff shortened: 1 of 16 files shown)


NadiaHG:
Can we also remove the exec_depend?  https://github.com/ipa320/cob_robots/blob/indigo_dev/cob_bringup/package.xml#L30

floweisshardt:
we cannot remove the exec_depend because we still have launch files which need cob_default_env_config, e.g. dashboard.launch

NadiaHG:
Yes, you are right.

CR: OK

PR #395, opened 2015-12-15
emergency_stop_state has to be a global topic

NadiaHG:


fmessmer:
why?
removing the `ns` will also put other topics into global namespace, e.g. https://github.com/ipa320/cob_driver/blob/indigo_dev/cob_relayboard/ros/src/cob_relayboard_node.cpp#L100-L102

NadiaHG:
You are talking about the  relayboard_node, and actually these other topics are already global topics: https://github.com/ipa320/cob_robots/blob/indigo_dev/cob_bringup/drivers/relayboard.launch#L12 
My pull request modify only the namespace for the relayboard_sim node, and makes it consistent with the real node topics.

What do you suggest remap the /emergency_stop_state topic?

fmessmer:
Ah, my bad....didn't see that you are changing the simulation case...
This indeed is a discrepancy between sim and hw....so if with your PR the three topics "emergency_stop_state", "power_board/state", "power_board/voltage" all use the same namespace both in sim and on hw, then :+1: for this PR

floweisshardt:
CR: OK

fmessmer:
Do not merge yet as there is a discussion going on in #403

fmessmer:
Conclusion is to merge this see https://github.com/ipa320/cob_robots/issues/403#issuecomment-179844477

PR #396, opened 2015-12-15
Updated base solo

NadiaHG:


NadiaHG:
needs https://github.com/ipa320/cob_robots/pull/395

NadiaHG:
test on cob4-2: OK

floweisshardt:
CR: OK

mgruhler:
Where is the stuff from `base_collision_observer.launch` now started?

floweisshardt:
this is only for the base_solo case which is a recovery mode if the complete robot bringup cannot be startet. Thus there's no need to launch the base_collision_observer here.
The base_collision_observer is still launched in the robot.xml files (complete bringup)

mgruhler:
Right, makes sense! Fine for me then.

fmessmer:
but why not keep it - at least commented out (as it is done for cob4's)?

fmessmer:
Ok, accepted...merging

Issue #348, opened 2015-11-04
consistent camera topic names (hw, sim)


floweisshardt:
taken from #347: 
cam3d topic names need a significant review anyway as topics in simulation do not match the topic names on real hardware for most of the kinect/softkinetic/prosilica sensors....

As #346 tries to introduce a consistent naming in simulation...can we tackle this issue after #346 has been merged and then try to remove trailing slashes for link_names wherever it applies?

fmessmer:
The `cob_common` part has been started in https://github.com/ipa320/cob_common/pull/179
...but it is not yet finalized....`cob_robots` is still missing...and wherever this change might have consequences (i.e. perception, moveit, rviz configs, ...)

fmessmer:
here is a comparison of the current topic from a cam3d in simulation and hw:
https://gist.github.com/ipa-fxm/d37e2e3f7e7ce9f8884f

fmessmer:
I already looks quite similar...still there are a few things, I (personally) would like to have clarified by some perception experts/users @ipa-rmb @ipa-josh 
1. In simulation, we have `depth/image_raw`, but `depth_registered/camera_info` and `depth_registered/points`. Shouldn't the `image_raw` topic always be in the same namespace as the respective `camera_info` topic?
2. Would you want to have all three topic within `depth` namespace or all three topics within `depth_registered` namespace or would you want all six topic?
3. What about the `ir/image`, `rgb/image_rect_color` topics?
4. Is it possible to compose a list of topics from the hw list that we actually do subscribe somewhere? I.e. if you are to choose topics from that list, which would you a) definitely need and b) definitely not need

Although, cob4's currently do not have stereo cameras...would it be possible to compose such a list for e.g. the prosilica cameras?

ipa-rmb:
Here are some answers to your questions from josh and me:
1. depth/image_raw is not used by josh or me. It would probably make more sense to rename it to the depth_registered namespace if the real hardware driver does the same.
2. We need all six variants, the three in depth name space and the other three from depth_registered.
3. Has not been used by us so far.
4. Usually the topics rgb/image (or rgb/image_raw or rgb/image_color -> depends on driver version) and rgb/camera_info and depth/camera_info, depth/points and depth_registered/camera_info, depth_registered/points are needed. All compressed topics are probably never used but they are created automatically anyways whenever a image_transport is used for image transmission. However, I see limited use of this list because the names change frequently in the original driver, so we are used to adapt topics anyways.

[...]

PR #398, opened 2015-12-15
review image_flip parameters

NadiaHG:


NadiaHG:
Test on cob4-2: OK

floweisshardt:
CR: OK

fmessmer:
Is this consistent with simulation?
Could you compare and create an issue, if it's not, in order to document the discrpancy.....might be related to https://github.com/ipa320/cob_robots/issues/348

floweisshardt:
this PR is OK. If there are differences between sim and real robot, we sould adapt simulation to fit the real robot.

fmessmer:
give me one more day to figure out what the situation in simulation is...

fmessmer:
- topic remaps look ok
- can't say much about the changes in the param files
- several things seem to have changed in the rviz confi...also the rviz configs for the raw and cob3 have not been adjusted...(they also use image_flip)

floweisshardt:
the rviz config files for cob3 and raw don't have image plugins configured, so no need to adapt image namespace here.

fmessmer:
Just one more thing, I found when looking at the topic list coming out of image_flip:
Wwhile the topics `_upright/depth_registered/points` and `_upright/rgb/image_raw` do exist,
there is no `_upright/depth_registered/camera_info` and no `_upright/rgb/camera_info` topic...

Is this a problem? Shouldn't there always be a `camera_info` topic in the same namespace as `image_raw` or `depth_registered/points`?

NadiaHG:
But the topics /$(arg camera_name)/depth_registered/camera_info and  /$(arg camera_name)/rgb/camera_info  are there?
The image flip is just used to visualise the image. The node has as input the original image and as output the same image but rotated. It has noting to do with the camera_info.

floweisshardt:
+1 for merging

fmessmer:
yes, also @ipa-rmb confirmed that not having the `camera_info` in the `upright`-namespace is not a problem...

cob_hardware_config/cob4-1/config/image_flip_params_sensorring_cam3d.yaml
```diff
@@ -1,30 +1,22 @@
 # rotation mode (0=fixed angle as defined  by rotation_angle, 1=automatic upright rotation against gravity using reference coordinate frame with upwards directed z-axis, 2=as 1 but image rotation only in 90deg steps -> faster)
 # int
-rotation_mode: 0
+rotation_mode: 1
```
ipa-rmb, line 3: Here you have activated the computationally most heavy option. Are you sure to do this on default? This rotates the image in arbitrary directions. Do you have a camera that is not built in accurately enough?

floweisshardt, line 3: good point, but I fear we need it that way because if we bend torso and/or head, we'll rotate the camera freely in space and thus need to use the gravity method.

ipa-rmb:
Ok fair enough. The computational burden is actually not that high (maybe <20 ms), waiting for tf is probably much longer.

PR #402, opened 2016-01-13
add scan unifier to simulated robots

floweisshardt:
necessary for using navigation from cob_navigation

### DIFF
--- a/cob_controller_configuration_gazebo/launch/robots/default_controllers_cob3-2.launch
+++ b/cob_controller_configuration_gazebo/launch/robots/default_controllers_cob3-2.launch
@@ -28,6 +28,9 @@
 		<arg name="robot" value="$(arg robot)"/>
 		<arg name="name" value="base_laser_top"/>
 	</include>
+	<include file="$(find cob_bringup)/drivers/scan_unifier.launch" >
+		<arg name="robot" value="$(arg robot)" />
+	</include>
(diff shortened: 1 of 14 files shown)


mgruhler:
Seems fine.

Just our of curiosity, for which modes do you need it, actually?
I think currently, the `cob_navigation_global` works on the relayed `scan_combined`  topic. Maybe not the SLAM one. Or am I wrong?

floweisshardt:
It was necessary for cob_navigation_local with DWA. I didn't test if cob_navigation_global and cob_navigation_slam would have worked without it. What's the difference between scan_combined and scan_unified? Where is scan_combined launched from (bringup or navigation launch files)?

mgruhler:
I did not find where this was required...
Do you know where the `scan_unified`  topic is used?

`scan_unified` is the output of the `scan_unifier`, i.e. one topic which is built out of mutltiple laser scan topics.

`scan_combined` is the relay topic of the respective laser scans. This WAS used in `amcl` and there were the respective relays.

Now, this has been changed to `scan_unified` there as well. Thus, the relays have been removed.
In `cob_navigation_local`, I don't see where `scan_unified` is used, but it is definitely in `cob_navigation_global` and `cob_mapping_slam`.

fmessmer:
LGTM

PR #397, opened 2015-12-15
[cob4-2] New torso pcs

NadiaHG:

PR #408, opened 2016-02-04
Pc monitor config symlinked

fmessmer:
this addresses my comment from #397 
@ipa-nhg can you please review it?

### DIFF
--- a/cob_bringup/robots/cob4-1.launch
+++ b/cob_bringup/robots/cob4-1.launch
@@ -2,12 +2,13 @@
 <launch>
 
 	<include file="$(find cob_bringup)/robots/cob4-1.xml" >
-		<arg name="cob4-1-b1" value="cob4-1-b1"/>
-		<arg name="cob4-1-t1" value="cob4-1-t1"/>
-		<arg name="cob4-1-t2" value="cob4-1-t2"/>
-		<arg name="cob4-1-t3" value="cob4-1-t3"/>
-		<arg name="cob4-1-s1" value="cob4-1-s1"/>
-		<arg name="cob4-1-h1" value="cob4-1-h1"/>
+		<!-- cob4-1 is currently a special config setup of cob4-2 hardware -->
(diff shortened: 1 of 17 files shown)


cob_bringup/robots/cob4-1.launch
```diff
@@ -2,12 +2,12 @@
 <launch>
 
 	<include file="$(find cob_bringup)/robots/cob4-1.xml" >
-		<arg name="cob4-1-b1" value="cob4-2-b1"/>
-		<arg name="cob4-1-t1" value="cob4-2-t1"/>
-		<arg name="cob4-1-t2" value="cob4-2-t2"/>
-		<arg name="cob4-1-t3" value="cob4-2-t3"/>
-		<arg name="cob4-1-s1" value="cob4-2-s1"/>
-		<arg name="cob4-1-h1" value="cob4-2-h1"/>
+		<arg name="cob4-1-b1" value="cob4-1-b1"/>
```
NadiaHG: I think that it is wrong, this value is also used to define the address(hostnames) of the machines https://github.com/ipa320/cob_robots/blob/indigo_dev/cob_bringup/robots/cob4-1.xml#L22

fmessmer:
you are right...I also added a comment, why we are doing this "magic" atm

NadiaHG:
Now the pc_monitor will not find the yaml configuration files, you have two options :
-  Rename the yaml files to ../con_hardware_config/cob4-**1**/config/pc_monitor_cob4-**2**-x1.yaml
- Don’t use the arg to set the pc value https://github.com/ipa320/cob_robots/blob/indigo_dev/cob_bringup/robots/cob4-1.xml#L27 


```
<arg name="pc" value="cob4-1-b1" />
```

fmessmer:
seems like I haven't thought the whole thing through....we do not seem to benefit from having symlinks here...it's ugly both ways...so we just keep your PR in https://github.com/ipa320/cob_robots/pull/397 and close this one...

PR #412, opened 2016-02-09
Parameter updates torso

fmessmer:
based on tests on cob4-1 hardware
results in acceptable motion (jerkyness-wise)...still the low update_rate introduces some lag


NadiaHG:
FR: OK

floweisshardt:
CR: OK
I've seen this on cob4-2 HW and looks good.

fmessmer:
Well merging this one, only makes sense together with the ones in cob_control...
However, I want to review/test those a bit more this afternoon and only merge them after that...

PR #410, opened 2016-02-04
[WIP] Feature cob4s with head - simulation only

fmessmer:
**Must not be merged until head is mounted to real HW!**

@ipa-fmw @ipa-tif @ipa-nhg FYI
This branch can be used for simulating/visualizing the cob4-1 and cob4-2 including the 3dof head!

The default head joint configurations seem not to be correct for the kinematic structure yet (as of https://github.com/ipa320/cob_common/issues/145)

PR #413, opened 2016-02-09
add 3dof head for cob4-1 within simulation only

fmessmer:
as requested


fmessmer:
@ipa-fmw please test in sim and on hardware (whether it breaks anything)

Also, I noticed that "head" is mentioned in the telop.yaml for cob4-2....maybe @ipa-nhg knows whether it can be removed safely...

cob_hardware_config/cob4-1/config/head_self_collision.yaml
```diff
@@ -0,0 +1 @@
+
```
NadiaHG, line 1: empty file? is it right?

cob_hardware_config/cob4-2/config/teleop.yaml
```diff
@@ -72,7 +72,7 @@ components: {
   head: {
     sss_default_target: "front",
     velocity_topic_name: "/head/joint_group_velocity_controller/command",
-    joint_velocity: [0.05, 0.05],
+    joint_velocity: [0,05, 0.05, 0.05],
```
NadiaHG: typo mistake:
0,05 --> 0.05

cob_default_robot_config/cob4-1/command_gui_buttons.yaml
```diff
@@ -32,6 +32,21 @@ group17:
...
   buttons: [[torso_front,trigger,torso_front],
     [torso_front_home,trigger,torso_front_home]]
+group18:
+  group_name: head
+  component_name: head
+  buttons: [[stop,trigger,stop],
+    [init,trigger,init],
+    [recover,trigger,recover],
+    [home,move,home],
+    [front,move,front],
```
NadiaHG, line 42: Are these positions already tested?

cob_hardware_config/cob4-1/config/head_self_collision.yaml
fmessmer, line 1: yes, empty file is fine for now...it's only used for twist_control with CollisionAvoidance enabled...

cob_default_robot_config/cob4-1/command_gui_buttons.yaml
fmessmer, line 42: no, they are not
see also my comment in https://github.com/ipa320/cob_robots/pull/410#issue-131418376

cob_hardware_config/cob4-2/config/teleop.yaml
fmessmer: well spotted...thanks!

fmessmer: @ipa-nhg 
why needs "head" to be included in teleop.yaml at all - for a robot that does not have a head?
can the tag be removed? what about all the parameters that have "head" in its name (e.g. `head_sss_default_target_button`)?

fmessmer:
@ipa-nhg @ipa-fmw we need to discuss about [this comment](https://github.com/ipa320/cob_robots/pull/413#discussion_r52428232)

floweisshardt:
tested on cob4-1: OK
tested in sim: OK

cob_hardware_config/cob4-2/config/teleop.yaml
NadiaHG: if a robot does't have a head , the component should not be included , for example for raw3-1: https://github.com/ipa320/cob_robots/blob/indigo_dev/cob_hardware_config/raw3-1/config/teleop.yaml 

Teleop will find all the components listed under`components: {` and allow all the modes defined (`joy_num_modes:` ) for all the listed components.

For example for cob4-2 ( https://github.com/ipa320/cob_robots/blob/indigo_dev/cob_hardware_config/cob4-2/config/teleop.yaml ) you have a list of 6 components (base, torso, head, sensorring, arm_left, arm_right), the parameter joy_num_modes is 4, and the button functions (parameters *_button) are also defined for 4 modes. 

that means that you can initialize:
- base, torso, head, sensorring, arm_left, arm_right

and you can move:
1. the base-twist command 
2. the torso, head, sensorring, arm_left and arm_right to a default position with a trajectory controller
3. the torso, head, sensorring, arm_left and arm_right  using the velocity controller
4. the torso, head, arm_left and arm_right using the twist controller.

For example for raw3-1 ( https://github.com/ipa320/cob_robots/blob/indigo_dev/cob_hardware_config/raw3-1/config/teleop.yaml ) you have a list of 1 components (base),  the parameter joy_num_modes is 1, and the button functions (parameters *_button) is only for mode1 defined:

that means that you can initialize:
- base

and you can move:
1. the base-twist command 

if the `head_sss_default_target_button` is defined and you didn't include "head" as component or you only have one mode , the parameter will be just not used

PR #414, opened 2016-02-10
add 3dof head for cob4-1 within simulation only

floweisshardt:
superceeds #413
- adds default_robot_behaviour to simulation
- fixes head poses


fmessmer:
so according to @ipa-nhg's [comment](https://github.com/ipa320/cob_robots/pull/413#discussion_r52475733), parameters with the name X_button where X is a component that does not exist for the respective robot are not needed....

this is also the case for e.g. raw's....they have minimall teleop.yaml with just the buttons needed for base

thus, one last question for this PR:
should we "clean up" the teleop.yaml for cob's so that parameters X_button are removed in case component X is not part of the robot?

I also added a bullet to issue https://github.com/ipa320/cob_command_tools/issues/112 regarding "component name"-independent parameter names...

floweisshardt:
+1 for cleaning up unused button parameters, but in a separate PR --> see issue #416
+1 for merging this one

NadiaHG:
FR: OK

PR #423, opened 2016-02-12
delete old and unused base velocity smoother config

floweisshardt:
is replaces by velocity_smoother.yaml using new smoother implementation


fmessmer:
FR ok

Should we put one of the yaml's into the cob_base_velocity_smoother` package as an example config?

mgruhler:
FR ok

+1 for putting one example yaml in the package.

NadiaHG:
FR: OK

floweisshardt:
-1 for putting an example config in the package
+1 for documenting all parameters on the wiki page http://wiki.ros.org/cob_base_velocity_smoother?distro=indigo

fmessmer:
Then we also would need to clarify that now there are two base_velocity_smoother implementations within the cob_base_velocity_smoother package...and explain their parameterization respectively!

fmessmer:
I did the according update on the wiki

PR #426, opened 2016-02-12
use softlinks to cob4-2 for most cob4-1 configs

floweisshardt:


fmessmer:
why don't we also symlink `base_driver.yaml`? (although it is mentioning cob4-1 this is just pointing to yet another symlink which is in common/ElmoSimplIQ.dcf)

NadiaHG:
+1 Symlink base_driver.yaml, sensorring_driver.yaml and torso_driver.yaml are already softlinks , using `cob4-2/config/Elmo.dcf` .

The softlink `cob4-1/config/Schunk_0_63.dcf` can be deleted, cob4-1 hasn't got any Schunk component, and if we use the `*_driver.yaml` files as links from cob4-2 , the `cob4-1/config/Elmo*.dcf` are also unnecessary.

fmessmer:
we should remove the `*.dcf` symlinks in the cob4-1 and cob4-2 (as long as they are symlinks!) and use a direct reference to `common/*dcf` in the driver.yaml....

fmessmer:
Let's hold this PR a bit longer until all the other PRs modifying the `cob_hardware_config` yaml's have been merged and stabilized...

fmessmer:
finalized together with @ipa-fmw directly...

PR #427, opened 2016-02-12
set lock priority for twist mux

floweisshardt:
should be higher than normal joystick priority (is 100)


fmessmer:
what's the effect of this?

floweisshardt:
the effect is that if you publish a std_msgs/Bool to the `joystick_lock` the twist mux will not forward any twist because `joystick_lock` has the highest priority.

We'll use that to "pause" the robot without aborting e.g. navigation (soft stop)

fmessmer:
as discussed with @ipa-fmw, the topics and priorities are now more intuitive and consistent...
@ipa-bnm FYI, in case you need to adjust these topic names in the state machine...!

PR #428, opened 2016-02-12
reduce deceleration factor in velocity smoother

floweisshardt:
this is used for cob4-1 too, as soon as we accept #426


NadiaHG:
FR: OK

PR #431, opened 2016-02-18
proper stop, init and recover via command_gui

fmessmer:
e.g. torso did not stop movement after pressing stop button on command_gui


fmessmer:
tested on cob4-1

floweisshardt:
CR: OK

PR #433, opened 2016-02-19
Fix cob41 bringup

fmessmer:
head pc currently not used - as long as Windows NUC is mounted

PR #434, opened 2016-02-19
rename canopen driver namespace

floweisshardt:
requested by @ipa-mdl and @ipa-fxm

### DIFF
--- a/cob_bringup/drivers/canopen_402.launch
+++ b/cob_bringup/drivers/canopen_402.launch
@@ -7,7 +7,7 @@
 	<arg name="can_device"/>
 
 	<!-- parameters for CANopen node -->
-	<node ns="$(arg component_name)" pkg="canopen_motor_node" type="canopen_motor_node" name="$(arg component_name)_driver" output="screen" clear_params="true">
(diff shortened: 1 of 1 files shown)


fmessmer:
this has effects on the diagnostics aggregation

fmessmer:
I cherry-picked and will fix it in #433

fmessmer:
I don't know why but the diagnostics aggregator is not able to deal with the diagnostics containing e.g. `torso driver` instead of `torso_driver`...whatever, I do (`contains`, `startswith`, `find_prefix` or `regex`) all the torso, base and other diagnostics are not mapped to the respective components anymore but can only be seen in "Others"....I vote for keeping the driver node name as `/component_name/componen_name_driver`....

mathias-luedtke:
some background: the generic analyzer checks on the raw name, including the `/`. Afterwards `/` has  to be substituted, because it is used as group separator in robot monitor (https://github.com/ros/diagnostics/blob/master/diagnostic_aggregator/include/diagnostic_aggregator/status_item.h#L53).

PR #435, opened 2016-02-19
restructure canopen driver yamls and remove canX yamls

### LINKED ISSUE #420: remove canX.yaml files and integrate into component specific yaml files
currently can0.yaml is used on the base pc for the base and on the torso pc for the arms/sensorring/torso. If we keep a common can0.yaml we cannot have different CAN configurations for e.g. base and torso.

@ipa-mdl had some suggestion how to restructure

fmessmer:
fixes #420 according to @ipa-mdl's concept

I did not adjust the driver.yaml's of:
- raw3-3
- raw3-6
  as they still use the old ipa_canopen config layout in there respective yaml file and we do not have access to the hardware....

### DIFF
--- a/cob_bringup/drivers/canopen_402.launch
+++ b/cob_bringup/drivers/canopen_402.launch
@@ -8,7 +8,7 @@
 
 	<!-- parameters for CANopen node -->
 	<node ns="$(arg component_name)" pkg="canopen_motor_node" type="canopen_motor_node" name="$(arg component_name)_driver" output="screen" clear_params="true">
-		<rosparam command="load" file="$(arg pkg_hardware_config)/$(arg robot)/config/$(arg can_device).yaml" />
+		<rosparam param="bus/device" subst_value="True">$(arg can_device)</rosparam>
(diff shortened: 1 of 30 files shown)


floweisshardt:
FR: OK

fmessmer:
After merging this, we should switch the `interval_ms` back from 10 to 20 in all `base_driver.yamls` in order to gain the control frequency of 50Hz...that has been used by the old undercarriage_ctrl driver (see [here](https://github.com/ipa320/cob_driver/blob/indigo_dev/cob_undercarriage_ctrl/ros/src/cob_undercarriage_ctrl.cpp#L135))

floweisshardt:
tested on cob4-2: OK (setting interval_ms to 20 seems to solve shaking behaviour)

fmessmer:
> (setting interval_ms to 20 seems to solve shaking behaviour)

See my [comment](https://github.com/ipa320/cob_robots/pull/441#issuecomment-188218048) in #441

PR #436, opened 2016-02-19
rename canopen node and adjust diagnostics

fmessmer:
as discussed with @ipa-mdl - follow-up of #434 
- renamed the node name
- updated `contains` tags of the canopen components in diagnostic_analyzer for robots using canopen components


floweisshardt:
FR: OK

PR #439, opened 2016-02-22
setting default parameters for backwards compatibility with configurable odometry_controller

fmessmer:
**only merge together with https://github.com/ipa320/cob_control/pull/90**

as the hardcoded default values of the odometry_controller have been changed in https://github.com/ipa320/cob_control/pull/90, this PR sets the according frames again in the yaml for backwards compatibility (until nav packages have been converted to new defaults).....this restores the behavior as it was before

### DIFF
--- a/cob_hardware_config/cob3-2/config/base_controller.yaml
+++ b/cob_hardware_config/cob3-2/config/base_controller.yaml
@@ -37,6 +37,10 @@ twist_controller:
 odometry_controller:
   type: cob_omni_drive_controller/OdometryController
   publish_rate: 50
+  frame_id: "/odom_combined"
+  child_frame_id: "/base_footprint"
+  cov_pose: 0.1
+  cov_twist: 0.1
   defaults: # default settings for all wheels, can per overwritten per wheel
     wheel_radius: 0.080 # Radius of the wheels in [m]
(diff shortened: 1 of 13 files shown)


cob_hardware_config/cob3-2/config/base_controller.yaml
```diff
@@ -37,6 +37,10 @@ twist_controller:
 odometry_controller:
   type: cob_omni_drive_controller/OdometryController
   publish_rate: 50
+  frame_id: "/odom_combined"
```
mgruhler, line 40: I guess we can get rid of the leading slashes here, as all calls to tf have the slash stripped under the hood, right?
Problems will probably only occur once we manipulate the respective ids manually.

Or not?

fmessmer, line 40: As this PR has "backwards compatibility" in its name, I'm simply making sure that the values are set to the exact same value as before - i.e. including `/`.
Getting rid of the leading slash should happen in a new PR...

mgruhler:
CR: ok then.

fmessmer:
Tested on cob4-1, able to set frames and cov, works as expected...

PR #440, opened 2016-02-23
add missing image_flip nodes to simulation

fmessmer:


cob_bringup/robots/cob3-2.xml
```diff
@@ -114,6 +114,7 @@
 		</include>
 		<!--include file="$(find cob_bringup)/drivers/image_flip_nodelet.launch" >
```
NadiaHG, line 115: why nodelet with the argument `start_manager` false?

floweisshardt:
CR: OK

cob_bringup/robots/cob3-9.xml
```diff
@@ -157,6 +157,7 @@
 		</include>
 		<!--include file="$(find cob_bringup)/drivers/image_flip_nodelet.launch" >
```
NadiaHG, line 158: why nodelet with the argument `start_manager` false?

NadiaHG:
@ipa-fmw and @ipa-fxm  , what about 2 reviews are needed before merge a pull request? Also travis didn't finished

cob_bringup/robots/cob3-9.xml
fmessmer, line 158: I just kept the arguments as they were used for the real robot...there they are commented out anyway....I don't know why `start_manager` was set to `false` for cob3-2 and cob3-9...but I did not want to change that....

fmessmer:
travis is green now :wink:

PR #441, opened 2016-02-24
set all base driver configs to interval_ms=20

floweisshardt:
follow up on #435


fmessmer:
Conducting the same input test case and comparing the resulting plot below to https://github.com/ipa320/cob_control/issues/91#issuecomment-188154604, I don't think this significantly improves the base driving behavior....

twist_odom_interval_20ms

Still I think it is the correct thing to do!

fmessmer:
is included in #445 
Closing in favor of #445

Issue #403, opened 2016-01-19
raw3-1 emergency state topic in wrong ns


stefandoerr:
in raw3-1.xml the relayboard.launch is started in simulation mode which results in a wrong ns of the /emergency_stop_state topic (which is relayboard/emergency_stop_state) and the robot can not be moved.

NadiaHG:
This PR https://github.com/ipa320/cob_robots/pull/395  solves the problem

fmessmer:
Even if #395 solves the namespace problem in simulation....why is this node run in simulation mode **on real hardware** anyways? There seems to be something wrong about this...

mgruhler:
I guess because the r@w does not have a relayboard. How is it handled on cob4?

fmessmer:
If raw3-1 does not have a relayboard, we should not have the robot to require a topic coming from this node in order to be able to operate it...
Which node requires a message on `/emergency_stop_state` on the raw3-1? @ipa-srd @ipa-nhg could you check this with `rostopic info`?

destogl:
> Even if #395 solves the namespace problem in simulation....why is this node run in simulation mode on real hardware anyways? There seems to be something wrong about this...

This is something I found out a year a go or so, but I was never really sure and had time to change it for all robots. For raw3-5 should be changed in #400 

> I guess because the r@w does not have a relayboard.

raw3-5 has? Or at least can publish this topics through phidgetboard. Probably I have to look at it again more closely.... Anyway I solved it for raw3-5...

mgruhler:
Yes, the `emergency_stop_state`  topic can be published from a phidget board.

destogl:
But than on all r@ws (if all have phidged board) this:

```
<include file="$(find cob_bringup)/drivers/relayboard.launch" >
        <arg name="sim" value="true" />
</include>
```

should be changed for this:

```
<include file="$(find cob_bringup)/drivers/phidgets.launch" >
        <arg name="robot" value="$(arg robot)" />
</include>
```

as I did in #400 

And of course we should take care that phidget board configuration is correct.

mgruhler:
True, but not all have a phidget board (actually, I'm not sure if there not only two that actually have one). And there are different versions. We have to check with our r@w's, at least the two that are still at IPA. For the rest, we need to keep it as is.

destogl:
Ok check it, at KIT we have phidgetboard, so for raw3-5 it can be also changed.

benmaidel:
[...]

PR #445, opened 2016-03-04
enable velocity sensor for um2 mode

### LINKED ISSUE #98: [cob4_base][ros_canopen][cob_elmo_homing] Debugging cob4_base performance
Initial assumption:
cob4-base seems to behave "jerkier" since recently. This _could_ be related to switching to `ros_canopen`, but is not guaranteed to be the reason as the behavior has not been monitored in such detail before...so no comparable plots are available.

Creating this issue to document the progress and the insights.
@ipa-mdl @ipa-fmw @ipa-tif @ipa-bnm

(Setup is latest `ros_canopen` (as of https://github.com/ros-industrial/ros_canopen/commit/e513abc12cea620ab32b058bc90252a0e10ada97) and latest "ipa320/indigo_dev" (as of today))

fmessmer:
fixes https://github.com/ipa320/cob_control/issues/98

includes #441 

currently only needed for cob4-1/cob4-2...not required for cob4-3/cob4-4/cob4-6...not yet applicable for cob3's and raw's (https://github.com/ipa320/cob_robots/issues/403)...

### DIFF
--- a/cob_hardware_config/cob4-2/config/base_driver.yaml
+++ b/cob_hardware_config/cob4-2/config/base_driver.yaml
@@ -1,5 +1,5 @@
 sync:
-  interval_ms: 10
+  interval_ms: 20
   overflow: 0
 heartbeat:
   rate: 20
@@ -9,6 +9,7 @@ defaults:
   eds_pkg: cob_hardware_config
   eds_file: "common/ElmoSimplIQ.dcf"
   dcf_overlay:
+    "606A": "1" # read velocity from velocity sensor
     "607F": "240000" # max profile velocity
     "6083": "1000000" # profile acceleration
     "6084": "1000000" # profile deceleration


fmessmer:
Tested on cob4-1 (see https://github.com/ipa320/cob_control/issues/98)

fmessmer:
@ipa-fmw @ipa-bnm 
I think you need to merge this PR for the demo on Monday/Tuesday as the `UM` mode is already set to `UM=2` on the robot....
(The robot is operable without merging, but you will not receive any velocities in the `/base/joint_states` ....this also leads to a zero twist on the `odometry` topic...I'm not sure whether navigation uses the twist....)

fmessmer:
Tested on cob4-1...merging because cob4-1 HW is already configured and requires it...

PR #447, opened 2016-03-17
MLR actual version

NadiaHG:
After some hardware changes the scanners and the flexisoft system are not connected to the base PC. The pull request is just the adaptation of the software to the actual hardware.


NadiaHG:
An extra topic is the urdf , I think that it makes not sense create a new base.urdf.xacro model, because we dont have the 3d model of the actual base. Also the simulation can be adapted, but I propose remove it. I don't think that we should support a simulation model that will not be used because the simulation  have nothing to do with the real robot.

cob_bringup/robots/cob4-3.xml
```diff
@@ -23,25 +23,6 @@
...
-			<arg name="robot" value="$(arg robot)" />
-			<arg name="name" value="base_laser_left" />
-		</include>
-		<include file="$(find cob_bringup)/drivers/sick_flexisoft.launch" >
-			<arg name="robot" value="$(arg robot)" />
-		</include>
-		<include file="$(find cob_bringup)/drivers/scan_unifier.launch" >
-			<arg name="robot" value="$(arg robot)" />
-		</include>
-
```
fmessmer, line 44: ~~So those platforms do not use the Sick scanners and no FlexiSoft?
What do they use instead?~~
Where are they connected to?

fmessmer:
As to the URDF, the difference about the base is that they do not use the white cover, right? Those are just a single mesh and could be made parameterizable via xacro-argument...

And what do you think needs to be changed about the simulation?

cob_bringup/robots/cob4-3.xml
NadiaHG, line 44: They have still scanners(only 2) and flexisoft but the devices are not connected to the base computer, they have their own system to manage the sick components

fmessmer:
After discussion with @ipa-nhg, I think the proposed PR is the best approach (minimal effort) to adjust to the updates....as long as it is not a problem for them to have wrong tf's from the (3) laser scanner (mounted directly to the base in our URDF)...I'd also keep the simulation...as removing them would mean more efforts....it's a nice looking base the way it is....:wink:

Can be merged after travis passed....

fmessmer:
Does anyone know why travis fails? @ipa-fmw @ipa-nhg

floweisshardt:
I agree to update config and keep URDF and simulation as it is (standard cob4 base).

Travis fails with 

```
[/home/travis/catkin_ws/src/cob_robots/cob_controller_configuration_gazebo/launch/robots/default_controllers_cob4-6.launch]:
    error loading <rosparam> tag: 
    file does not exist [/home/travis/catkin_ws/src/cob_robots/cob_hardware_config/cob4-6/config/base_laser_front.yaml]
```

we're using the laser*.yaml files to configure laser filters in simulation thus we sould not remove them :-)

NadiaHG:
I re-added the configuration files and now travis is happy

floweisshardt:
FR: OK

fmessmer:
Fine with me

PR #451, opened 2016-04-01
Deleted unused parameter

NadiaHG:
roslaunch returns with error because the parameter is defined and not set


floweisshardt:
CR: OK

PR #455, opened 2016-04-04
add collision_monitor launch file

fmessmer:
This launch file starts the `collision_monitor` node that publishes  the `state_valid` topic based on MoveIt! planning scene that is used by `cob_safety_controller` to stop movements in case of collision.

We might want to add this launch file to all our robots (that use the `cob_safety_controller` package and have a valid `cob_moveit_config`)

Note that this introduces a dependency from `cob_bringup` to `cob_manipulation`, but I guess this is not a problem....as `cob_manipulation` is up-to-date and released...

This launch file might also later be used to add the `cob_obstacle_distance_moveit` node that replaces the component-specific `cob_obstacle_distance` nodes (see https://github.com/ipa320/cob_control/issues/101)

@ipa-mdl


floweisshardt:
if this introduces a dependency to cob_manipulation, we should add cob_manipulation in the metapackage

fmessmer:
no, dependencies are already specified accordingly in the `package.xml`

fmessmer:
I decided to rename the launch file and already added them to the "affected" robots

fmessmer:
mergin in order to allow @ipa-fxm-cm and @ipa-bfb a cleaner start for their contribution

PR #456, opened 2016-04-04
prepare using robots with cartesian controller

fmessmer:
this brings in according launch files and parameters for using twist_control with all cob4-1/cob4-2 components currently available....(just need to be uncommented)


floweisshardt:
FR: OK

fmessmer:
mergin in order to allow @ipa-fxm-cm and @ipa-bfb a cleaner start for their contribution

PR #457, opened 2016-04-04
Fix/pc monitor

floweisshardt:
config was missing for h32 pc


fmessmer:
needs to be tested with @ipa-nhg's latest head-pc setup

fmessmer:
rebased in #459...thus, closing

PR #449, opened 2016-03-22
harmonize base controller settings

floweisshardt:
this setup was tested on cob4-2 during store tests in ingolstadt
- unifies joint sequence
- applies controller parameters to all robots

PR #458, opened 2016-04-04
use base_controller values from ini file

fmessmer:
separated from #449 

needs to be tested on hardware
@ipa-fmw


floweisshardt:
FR: OK
we'll test this today on the robot

floweisshardt:
tested on cob4-1: OK

PR #459, opened 2016-04-04
add pc monitor config for h32

fmessmer:
rebase of #457 

needs to be tested with @ipa-nhg's latest head-pc setup

### DIFF
--- a/cob_hardware_config/cob4-1/config/pc_monitor_cob4-2-h32.yaml
+++ b/cob_hardware_config/cob4-1/config/pc_monitor_cob4-2-h32.yaml
@@ -1,0 +1,1 @@
+../../cob4-2/config/pc_monitor_cob4-2-h32.yaml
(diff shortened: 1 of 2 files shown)


fmessmer:
The head-pc setup is still not working right:
- [ ] accounts need to be generated for all users that have an account on the base-pc
- [ ] perform `ssh-copy-id` and make sure there is a valid `.bashrc` (cob_default.bashrc is fine) that is sourced

Also, we have an issue with the `cob_bringup/env.sh` does have an overlay of `cob_robots` on the base-pc **but not on the head-pc** because the machine tag passes as [env-loader](https://github.com/ipa320/cob_robots/blob/indigo_dev/cob_bringup/robots/cob4-1.xml#L199) the overlay path - which does not exist on the blank head-workspace....

fmessmer:
Discussion on  the env-loader script is moved to a new issue #461 and a proposed WIP-solution can be found in #462

floweisshardt:
for this PR: CR: OK
we should merge this and continue discussion in #461

Issue #461, opened 2016-04-07
[cob_bringup] env.sh as env-loader script


fmessmer:
@ipa-nhg @ipa-fmw

This is evolving from https://github.com/ipa320/cob_robots/pull/459#issuecomment-205390813

Beside the discussion about what to do about `env.sh`, the following things need to be done (@ipa-nhg)
- [ ] accounts need to be generated for all users that have an account on the base-pc
- [ ] perform `ssh-copy-id` and make sure there is a valid `.bashrc` (cob_default.bashrc is fine) that is sourced

fmessmer:
The problem is that we cannot use `$(find cob_bringup)` within the `env-loader` part of the `machine`-tag as it gets resolved **before** being passed to the remote PC!

Thus, if an overlay of `cob_robots` exists on the b1-PC, "/u/user-name/git/care-o-bot/src/cob_robots/cob_bringup/env.sh" is passed over to the remote PC, e.g. h32. 
However, if no overlay exists on the remote-PC, this file does not exist, and `cob_bringup/robot.launch` will fail...

In #462, I tried to make the env-loader-script a launch argument and tried to use `~/.bashrc` which is a relative path and could be resolved to the remote-PCs local `.bashrc` which can be different from the b1-PC `.bashrc`....

However, according to [here](http://wiki.ros.org/roslaunch/XML/machine) the env-script needs to be a shell script. (Passing `~/.bashrc` fails with

```
fxm@cob4-2-b1:~$ roslaunch cob_bringup robot.launch robot:=cob4-1
... logging to /u/fxm/.ros/log/eb99b746-fc8e-11e5-b1a9-eca86bff61c1/roslaunch-cob4-2-b1-9337.log
Checking log directory for disk usage. This may take awhile.
Press Ctrl-C to interrupt
WARNING: disk usage in log directory [/u/fxm/.ros/log] is over 1GB.
It's recommended that you use the 'rosclean' command.

started roslaunch server http://cob4-2-b1:46077/
remote[cob4-2-t2-0] starting roslaunch
remote[cob4-2-t2-0]: creating ssh connection to cob4-2-t2:22
[...]

PR #462, opened 2016-04-07
[WIP] fix/env loader script

fmessmer:
this is an approach to solve #461 
(it includes #459)

Solution is still not ideal and needs more discussion
@ipa-nhg @ipa-fmw


NadiaHG:
Not required because cob4-2-h1 PC will be 64 bits soon.

PR #471, opened 2016-04-28
Test Head 3dof

NadiaHG:


cob_default_robot_config/cob4-2/head_joint_configurations.yaml
```diff
@@ -5,16 +5,16 @@ default_vel: 0.4
 
 home: [[0.0, 0.0, 0.0]]
 
-front: [[0.0, 0.0, 0.0]]
-front_down: [[1.57, -3.14, 1.57]]
-front_down_twisted: [[1.57, -3.14, -1.57]]
+front: [[0.0, 0.0, 3.14]]
```
floweisshardt, line 8: why are those poses different to simulation? @ipa-fxm updated the poses for cob4-1 in sim some weeks ago.
If this is correct for the hardware, we should double check the URDF and simulation model (especially turining direction and reference positions of axis)

NadiaHG, line 8: I also notified this error problem, the urdf has to be updated. The real hardware doesn't correspond to the current urdf model. Please keep this pull request open until we solve the problem.

NadiaHG, line 8: The problem is solved after update the URDF model , see https://github.com/ipa320/cob_common/pull/189 
These changes are already undo https://github.com/ipa320/cob_robots/pull/473/commits/d25d82cf03d900d2f59e92d08f82e0a0dcbeaac0 on the https://github.com/ipa320/cob_robots/pull/473

NadiaHG:
The pull request can be closed, the https://github.com/ipa320/cob_robots/pull/473 already contains the changes

PR #478, opened 2016-05-24
Remove dependency to cob_obstacle_distance_moveit

NadiaHG:


NadiaHG:
As discussed: https://github.com/ipa320/cob_robots/pull/468

mathias-luedtke:
not sure if we should remove it if `cob_twist_controller` uses it.
@ipa-fxm-cm: what happens if the controller is run without the distance monitor per default?

MaChristoph:
@ipa-mdl : If the twist_controller is started without the distance informations then the collision avoidance won't work. The cob_obstacle_distance_moveit node is the "new" one and replaces the "old" cob_obstacle_distance node, since it's computation time is much more efficient. So you might delete the cob_obstacle_distance dependency instead of the cob_obstacle_distance_moveit dep.

floweisshardt:
@ipa-nhg: what's the status here? can this be merged?

floweisshardt:
CR: OK, merge conflicts solved in https://github.com/ipa320/cob_robots/pull/517

PR #479, opened 2016-05-24
Add SRDF to cob_hardware_config

NadiaHG:


NadiaHG:
As discussed: https://github.com/ipa320/cob_robots/pull/468

fmessmer:
All SRDFs in cob_mobeit_comfig have been updated and tested before I left...
You can use them rather than empty!
Only need to update in case URDF chamged (3dof head!)

mathias-luedtke:
We agreed to use empty SRDFs for now.
We will migrate the configured  SRDFs later, after a more detailed check.
(The collision monitor detects false collisions occasionally.)

CR: OK

PR #482, opened 2016-06-08
never allow collissions for base/torso and torso/head

floweisshardt:
prevents the collission_monitor to send `state=false` due to colliding meshes. needed for head movements of cob4-1.


NadiaHG:
CR: OK

PR #484, opened 2016-06-21
use dummy state publisher instead of real collission monitor

floweisshardt:
current setup with collission monitor is not working reliably yet. Sometimes there is a `is_valid=False` which causes an emergency stop.

This PR should be reverted as soons as the collission monitor is working fine.

tested on cob4-1/cob4-2: OK


NadiaHG:
Travis finishes with a failure:

`Missing package dependencies: cob_bringup/package.xml: rostopic`

floweisshardt:
travis is good. new try...

NadiaHG:
Travis is now happy 
CR: OK

PR #486, opened 2016-06-22
include base collision observer

NadiaHG:


floweisshardt:
FR: OK

NadiaHG:
Test on cob4-1: OK

PR #488, opened 2016-06-23
disabled battery monitor sound/light and emmonitor sound

benmaidel:
disabled battery_monitors sound + light and emergency monitors sound output.
Done because the bms is always reporting low battery and in case of msh scenario we do not like to have sound output if em is active.


NadiaHG:
CR : Ok
Test on cob4-1: OK

floweisshardt:
CR: OK

PR #490, opened 2016-06-24
3d camera performance improvements

benmaidel:
- Start image flip inside the same nodelet manager as the camera is running. Otherwise using nodelets is nonsense
- inside openni2.launch was a param `data_skip` with the value 5. This means that only every fifths data package is used. Can somebody explain me why this was set? The result of this was a frame rate of 5-6 fps. Now we achieve 30 fps!

### DIFF
--- a/cob_bringup/drivers/openni2.launch
+++ b/cob_bringup/drivers/openni2.launch
@@ -5,7 +5,7 @@
 	<arg name="name" default="cam3d"/>
 	<arg name="device_id" default="#1"/>
 
-	<param name="/$(arg name)/driver/data_skip" type="int" value="5"/>
+	<param name="/$(arg name)/driver/data_skip" type="int" value="0"/>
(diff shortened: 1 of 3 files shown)


benmaidel:
@ipa-fmw @ipa-rmb @ipa-nhg 

do you know why data_skip was set to 5?

ipa-rmb:
I did not set this, and I was not aware of such parameter until now. Maybe it was used to limit processing load on the sensor PC?

ipa-rmb:
Or someone just hit the wrong line (image_modes are also 5)

NadiaHG:
@ipa-bnm no idea :( 
But thank you for find the error, can you please modify also the cob4-2.xml file? https://github.com/ipa320/cob_robots/blob/indigo_dev/cob_bringup/robots/cob4-2.xml#L198

Thanks!

benmaidel:
@ipa-nhg done

NadiaHG:
CR: OK

PR #491, opened 2016-06-29
Positions for head and torso

timhappy:


benmaidel:
Because this is project related, we moved it to msh.
We can close this PR

floweisshardt:
makes sense to move to msh repository

PR #492, opened 2016-07-02
params for battery capacity calculation added

timhappy:
requires ipa320/cob_driver#278


mgruhler:
CR: ok

benmaidel:
CR: ok

floweisshardt:
I'd like to avoid adding this parameters at all. Even if we follow up on the phidget solution we should be able to calculate a relative_ramaining_capacity in percent, so no need to get down to the voltage level here.

floweisshardt:
this PR is not relevant any more because we're using the phidget based power state publisher which calculates a relative_remaining_capacity

PR #493, opened 2016-07-03
respawn bms driver

benmaidel:


mgruhler:
CR: ok

why is this required, though?

benmaidel:
In msh we wrote a note to kill the bms driver if he is't receiving any more data over can. To make the bms restart after we killed him, we need to set this!

floweisshardt:
CR: OK

PR #495, opened 2016-07-12
phidget config for cob4-1

benmaidel:


floweisshardt:
modification should be made for cob4-2 and keep the softlink from cob4-1 to cob4-2.

benmaidel:
done!

floweisshardt:
CR: OK

PR #496, opened 2016-07-12
powerstate from phidget

benmaidel:
get powerstate informations from phidget board as long as the bms system is buggy

depends on:
- https://github.com/ipa320/cob_driver/pull/279

PR #498, opened 2016-07-14
Feature/powerstate from phidget

floweisshardt:
includes https://github.com/ipa320/cob_robots/pull/496

adds rosinstall dep to satisfy travis


benmaidel:
CR: ok

PR #509, opened 2016-07-22
Fix/disable head and sensorring for cob4-2

floweisshardt:
due to not working properly

includes https://github.com/ipa320/cob_robots/pull/510


NadiaHG:
FR : OK

PR #513, opened 2016-08-03
Updated cob4-3 config

mathias-luedtke:
- ignore BMS in diagnostics aggregator (**should be reverted later**)
- base calibration
- read currents and include them in joint_states
- stuck detector (needs https://github.com/ipa320/cob_control/pull/111)
- **do not respawn BMS driver**

### DIFF
--- a/cob_bringup/robots/cob4-3.xml
+++ b/cob_bringup/robots/cob4-3.xml
@@ -31,6 +31,10 @@
 		<include file="$(find cob_bringup)/controllers/base_controller_plugin.launch">
 			<arg name="robot" value="$(arg robot)"/>
 		</include>
+		<include file="$(find cob_omni_drive_controller)/launch/stuck_detector.launch">
+			<arg name="timeout" default="2.0" />
+			<arg name="threshold" default="0.174533" /> 
+		</include>
 
 		<!-- start additional packages -->
 		<include file="$(find cob_bringup)/tools/diagnostics_aggregator.launch" >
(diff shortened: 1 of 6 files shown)


cob_bringup/robots/cob4-3.xml
```diff
@@ -31,6 +31,10 @@
 		<include file="$(find cob_bringup)/controllers/base_controller_plugin.launch">
 			<arg name="robot" value="$(arg robot)"/>
 		</include>
+		<include file="$(find cob_omni_drive_controller)/launch/stuck_detector.launch">
```
NadiaHG, line 34: please add the `cob_omni_drive_controller` dependency  to the package.xml

floweisshardt:
https://github.com/ipa320/cob_control/pull/111#issuecomment-235715133 says that we can handle the stuck detection on the elmo firmware, which I prefer over handling it in ROS.
So please remove all stuck detection related modifications from this PR.

cob_bringup/robots/cob4-3.xml
mathias-luedtke, line 34: I wasn't aware that cob_omni_drive_controller was pulled in via cob_hardware_config. Made it explicit now.

mathias-luedtke:
> ipa320/cob_control#111 (comment) says that we can handle the stuck detection on the elmo firmware, which I prefer over handling it in ROS.

Me too, but it did not work out. Elmos can detect stuck errors (in velocity mode) only by analyzing the current. The high level controller can verify against the actual collision target.

floweisshardt:
FR: OK

NadiaHG:
FR: OK

PR #517, opened 2016-08-10
Remove dependency to cob_obstacle_distance_moveit

floweisshardt:
same as https://github.com/ipa320/cob_robots/pull/478 but merge conflicts resolved


NadiaHG:
CR: OK

(@ipa-fmw thanks!)

fmessmer:
I don't like this...just saying...
Also the moveit_config restructuring seems to be on hold...no need for motion_planning and/or collision_checking/monitoring with cob's atm?

floweisshardt:
there would be use cases, but unfortunately we're lacking competences until you're back!

PR #523, opened 2016-08-17
fix station docking for multiple stations

floweisshardt:


cob_hardware_config/cob4-2/config/docking_linear_nav.yaml
```diff
@@ -16,7 +16,7 @@ goal_threshold_rot: 0.08 #0.08
 speed_threshold_rot: 0.01 #0.08
 
 #tf parameters
-global_frame: station_reflector_link #/map
+global_frame: odom_combined #/map
```
mgruhler, line 19: do you really want to change this? `odom_combined` is not stationary wrt. `map`. Thus, in some edge cases (high slippage, ...) this will actually lead to a very bad performance...

floweisshardt, line 19: I just struggled over this because station_reflector_link does **not** even exist! I changed it to a frame which is available on all robots even without running navigation. In my tests there was no difference in setting this frame to `base_link`, `odom_combined` or `abs_xxx_123`. @ipa-mig: Do you know when this frame really makes a difference?

mgruhler, line 19: I guess it only makes a difference when there is a chance to accumulate larger odometry errors that get corrected via localization.

Thus, if the goal you try to move to via linear nav is relative to `odom_combined`, this should be fine. Also, if the tf `odom_combined->station_reflector_link` (through whatever chain) is fixed, then for this case there is no difference at all.

The only real difference will probably be when running with bad odometry, over a larger distance, targeting a fixed frame in `map`.

I don't know how this is actually working, thus, I cannot say ;-)

floweisshardt, line 19: ok, so good to merge? @ipa-mig?

mgruhler:
FR: ok

@ipa-srd?

stefandoerr:
FR: ok!

PR #527, opened 2016-08-23
add data skip launch argument for openni2 to limit CPU load

floweisshardt:

PR #529, opened 2016-08-23
increase error rate for hz monitor

floweisshardt:


floweisshardt:
merge together with #527

benmaidel:
CR: ok

NadiaHG:
CR: ok

PR #528, opened 2016-08-23
increase velocity thresholds for safety fields

floweisshardt:

PR #530, opened 2016-08-23
Fix/torso cam3d down

floweisshardt:
includes https://github.com/ipa320/cob_robots/pull/529
includes https://github.com/ipa320/cob_robots/pull/528


benmaidel:
CR: ok

NadiaHG:
CR: ok

PR #549, opened 2016-10-13
Cob3-6 twist controller configuration

bbrito:
Merge this pull request with this one: https://github.com/ipa320/schunk_modular_robotics/pull/176


NadiaHG (approved):
CR: OK

fmessmer:
finalized in #564 - the self-collision-links were not correct for cob3-6...
As this feature is only beta, it's ok to have an empty file

PR #555, opened 2016-11-15
Feature/simulation faker

HannesBachter:
Fake monitoring to have a valid robot_status in simulation

=> needs to be tested on hardware


cob_default_robot_config/cob4-2/torso_joint_configurations.yaml
```diff
@@ -11,7 +11,9 @@ front_up_full: [[3.14, -1.57]]
 front_down: [[-1.0, 0.5]]
 front_down_full: [[-3.14, 1.57]]
 
-back: [[0.0, 3.14]]
+back: [back_left]
```
benmaidel, line 14: I'm not sure if this will work. Have you tested this?

cob_controller_configuration_gazebo/launch/robots/default_controllers_cob4-1.launch
```diff
@@ -36,6 +36,8 @@
 		<arg name="sim" value="true"/>
 	</include>
 
+        <node pkg="rostopic" type="rostopic" args="pub /safety_controller/state_is_valid std_msgs/Bool 'data: true' -r10" name="fake_collission_monitor" output="screen"/>
```
benmaidel: - indentation
- why do we need this? Can't we just run the cob_collision_monitor in simulation?

benmaidel (changes requested):

cob_default_robot_config/cob4-2/torso_joint_configurations.yaml
HannesBachter: yes, works in simulation

cob_controller_configuration_gazebo/launch/robots/default_controllers_cob4-1.launch
floweisshardt: we're running the same fake on the real robot because the collision monitor caused troubles which are still not resolved. Maybe it is worth a try and investigating again.

floweisshardt (approved):
CR: OK (except for the indentation issue already raised by @ipa-bnm)

benmaidel (approved):

benmaidel (approved):

ipa-bnm:
@ipa-fmw do you know why there are still statuses to pass?

PR #561, opened 2016-12-08
coupled sensorring and head in one instance

mathias-luedtke:
merged sensorring config into head config, added remaps and relays.
needs https://github.com/ipa320/cob_command_tools/pull/164

PR #566, opened 2016-12-09
coupled sensorring and head in one instance with helper

mathias-luedtke:
supersedes #561

not yet tested on real hardware, but controllers can be loaded as expected.

Pros:
* does not mess with the controller config files (works nice with gazebo)
* rqt_joint_controller_manager works

Cons:
* slave controllers have absolute names
* slave controllers cannot be loaded with rqt_controller_manager


mathias-luedtke:
I have implemented another sync mechanism (https://github.com/ros-industrial/ros_canopen/pull/201) that might mix fix the problem without these hacks.

Issue #437, opened 2016-02-19
[cob_hardware_config] further simplify configuration structure (modules and components)


fmessmer:
The structure already got much simpler but there still is some potential.

I'm creating this issue for discussion following
- https://github.com/ipa320/cob_common/pull/175#issuecomment-154128294
- https://github.com/ipa320/cob_robots/pull/359

This has minor priority for now...

@fmessmer FYI

fmessmer:
After cleaning up unsupported robots #588, we will start introducing a more generic robot-/module-/component-base structure for all URDF, launch and config files... as discussed with @ipa-fmw @ipa-nhg @ipa-mdl 

This will allow further de-duplication of code, more consistency between robots, better harmonization of HW/Sim-behavior as well as easier future maintenance and up-scaling...

fmessmer:
#622 is a significant step towards a simplified configuration structure

I will apply a similar concept for the other "unifiable" components, e.g. torso, head, sensorring, arm(s),...

fmessmer:
Also, we would need to adjust the structure to the robot names we are using now to refer to a (physically existing) robot composition, i.e. `cob4-paul-ingolstadt`...following @ipa-tif's spreadsheet Verbauliste.xlsx

The ultimate goal could be to even generate launch files based on a module configuration using a tool based on @ipa-mdl's [roslaunch-dump](https://gist.github.com/ipa-mdl/66b14a895e13003a2f76e483bd00aa63)

fmessmer:
@ipa-mdl asked for input about `roslaunch-dump` (with the final goal being added to `roslaunch`)
https://github.com/ros/ros_comm/issues/1022

floweisshardt:
reminder for @ipa-fxm: extract sick scanners from base urdf + then remove nav configs for mlr bases as they do not have laser scanners anymore

fmessmer:
> extract sick scanners from base urdf + then remove nav configs for mlr bases as they do not have laser scanners anymore

done in https://github.com/ipa320/cob_robots/pull/654

fmessmer:
atm none of the solutions proposed by @ipa-fxm and @ipa-nhg fully meet our requirements
also current efforts require a final decision regarding future maintaining responsibilities and supported hardware

thus, I'm moving it to Backlog

floweisshardt:
we won't work on this for now

mathias-luedtke:
> we won't work on this for now

No need to close it.

fmessmer:
[...]

PR #571, opened 2016-12-15
[DoNotMerge] concept for unification of hw/sim-independent bringup nodes

fmessmer:
__This PR is (currently) not meant to be merged!__

I opened it to throw in a concept to be discussed with which we could achieve a further unification/harmonization between ```cob_bringup``` (i.e. hardware) and ```cob_bringup_sim``` (simulation).

The main idea is to extract common parts of ```cob_bringup/cobX-X.xml``` and ```default_controllers_cobX-X.launch``` (i.e. parts that occur in both of these files) into a launch file that - __identically__ - will be used by both cob_bringup and cob_bringup_sim!

What's left in ```cob_bringup/cobX-X.xml``` is hardware-specific (i.e. drivers) and will be named ```cob_bringup/cobX-X_hw.xml```. What's left in ```default_controllers_cobX-X.launch``` is simulation-specific and could be moved into a launch file in cob_bringup_sim (creating a similar structure analog to cob_bringup using cobX-X sub-folders).

In the end, the package ```cob_controller_configuration_gazebo``` will become obsolete and could be removed from cob_robots (the remaining scripts will be moved to cob_simulation).

Goal of this splitting would be to move as much into the common cobX-X.xml file which would guarantee hw-/sim-behavior (i.e. topics, actions, services) as similar as possible. Also, we would not need to remember to adjust simulation when adding something in hardware and vice-versa...

The whole effort can be seen as part of #437

@ipa-fmw @ipa-nhg @ipa-mdl @ipa-mig @ipa-bnm what do you think?
Is the intended concept clear? Don't hesitate to ask questions and discuss!

In the PR I only commented parts instead of removing.
What is commented in cobX-X_hw.xml is hw-independent and can be moved to cobX-X.xml.
What is commented in cobX-X.xml is hw-specific and stays in cobX-X_hw.xml

### DIFF
--- a/cob_bringup/robots/cob4-1.launch
+++ b/cob_bringup/robots/cob4-1.launch
@@ -3,6 +3,19 @@
 
 	<arg name="env-script" default="$(find cob_bringup)/env.sh"/>
 
+	<!-- start hw-specific nodes, i.e. drivers -->
+	<include file="$(find cob_bringup)/robots/cob4-1_hw.xml" >
+		<!-- cob4-1 is currently a special config setup of cob4-2 hardware -->
+		<arg name="cob4-1-b1" value="cob4-2-b1"/>
+		<arg name="cob4-1-t1" value="cob4-2-t1"/>
+		<arg name="cob4-1-t2" value="cob4-2-t2"/>
+		<arg name="cob4-1-t3" value="cob4-2-t3"/>
+		<arg name="cob4-1-s1" value="cob4-2-s1"/>
+		<arg name="cob4-1-h1" value="cob4-2-h32"/>
+		<arg name="env-script" value="$(arg env-script)"/>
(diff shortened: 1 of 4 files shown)


fmessmer:
I like https://github.com/ipa320/cob_robots/pull/583 better...

mgruhler:
even though this seems to be the more intuitive way to go, I also think that #583 is better.

This is mainly due to the fact that we can then use the exact same launch files and [don't stumble upon "hidden" nodes](https://github.com/ipa320/cob_robots/pull/571/files#diff-a006658e7a9b053d8419aee65cf163ceR27), that are actually not HW specific, but where it makes sense to have them in the "hw"-launchfile.

fmessmer:
closing in favour of #583

Issue #577, opened 2017-01-11
Use CMake-Macro to retrieve sub-dir-list


fmessmer:
use for ```robotlist``` and ```envlist``` e.g. for install tags in various packages

See http://stackoverflow.com/questions/7787823/cmake-how-to-get-the-name-of-all-subdirectories-of-a-directory

as suggested by @ipa-mdl in https://github.com/ipa320/cob_robots/pull/573#discussion_r93017007

mathias-luedtke:
File globbing might stop CMake from rebuilding as it does not detect changes..
I think we can export the (hand-crafted) robotlist from ` cob_hardware_config`.
In order to use it, we have to `build_depend` on `cob_hardware_config`.

In `cob_moveit_config` we should install files from the package directory with exclusion rules.

fmessmer:
exported lists are used/proposed in all related packages....

PR #579, opened 2017-01-11
Export robotlist

mathias-luedtke:
addresses #577

### DIFF
--- a/cob_default_robot_config/package.xml
+++ b/cob_default_robot_config/package.xml
@@ -15,4 +15,6 @@
 
   <depend>roslaunch</depend>
 
+  <build_depend>cob_supported_robots</build_depend>
+
 </package>
--- a/cob_hardware_config/package.xml
+++ b/cob_hardware_config/package.xml
@@ -16,6 +16,8 @@
   <depend>roslaunch</depend>
   <depend>rostest</depend>
 
+  <build_depend>cob_supported_robots</build_depend>
+  
   <exec_depend>cob_calibration_data</exec_depend>
   <exec_depend>cob_description</exec_depend>
   <exec_depend>cob_omni_drive_controller</exec_depend>
--- a/cob_bringup/CMakeLists.txt
+++ b/cob_bringup/CMakeLists.txt
@@ -1,33 +1,13 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(cob_bringup)
 
-find_package(catkin REQUIRED COMPONENTS roslaunch)
+find_package(catkin REQUIRED COMPONENTS cob_supported_robots roslaunch)
 
 catkin_package()
 
-set(robotlist
-  cob3-2
-  cob3-6
-  cob3-9
-  cob4-1
-  cob4-2
-  cob4-3
-  cob4-4
-  cob4-5
-  cob4-6
-  cob4-7
-  cob4-10
-  raw3-1
-  raw3-2
-  raw3-3
-  raw3-4
-  raw3-5
-  raw3-6
-)
-
 ### TEST ###
 if(CATKIN_ENABLE_TESTING)
-  foreach(robot ${robotlist})
+  foreach(robot ${cob_supported_robots_ROBOTLIST})
     message("testing for robot: ${robot}")
     roslaunch_add_file_check(robots/${robot}.xml)
   endforeach()
--- a/cob_bringup/CMakeLists.txt
+++ b/cob_bringup/CMakeLists.txt
@@ -1,12 +1,14 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(cob_bringup)
 
-find_package(catkin REQUIRED COMPONENTS cob_supported_robots roslaunch)
+find_package(catkin REQUIRED)
 
 catkin_package()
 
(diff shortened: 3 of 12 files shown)


mathias-luedtke:
should be used in `cob_calibration_data`as well.

fmessmer:
Wow, sweet....that will help keep things consistent quite a lot!

We should do the same for ```ENVLIST``` in ```cob_default_env_config``` and then use ```cob_hardware_config_ROBOTLIST``` and ```cob_default_env_config_ENVLIST``` wherever it is used!

mathias-luedtke:
> wherever it is used!

this is tricky as not all robots are configured for high-level packages.

fmessmer:
Sorry, I introduced a merge conflict when merging #578 
But I can take over and try to apply the new cob_hardware_config_ROBOLIST to further repositories....

However, when using this for ```cob_calibration_data```, we are introducing a cyclic dependency...as ```cob_calibration_data``` is used within ```cob_hardware_config```

mathias-luedtke:
> Sorry, I introduced a merge conflict

rebased

> However, when using this for cob_calibration_data, we are introducing a cyclic dependency.

New package `cob_supported_robots`?
Or keep list in  `cob_calibration_data`?

fmessmer:
> rebased

Also already did that in #582 

 > New package `cob_supported_robots`?
Or keep list in  `cob_calibration_data`?

I would keep the ```robotlist``` in ```cob_calibration_data``` as it is very unlikely that we forget to configure this for a new robot :wink:

mathias-luedtke:
> I would keep the robotlist in cob_calibration_data

This even reduces the dependencies (just `xacro`). I wasn't aware that `cob_hardware_config` depends on so many packages.

A `cob_supported_robots`  package would not have any dependencies apart from `catkin`..

floweisshardt:
nice solution to mimimise maintenance effort. But I don't like the fact that we introduce dependencies to cob_hardware_config from every package (also simulation related and possibly high-level packages). 

So I vote for introducing a new package `cob_supported_robots`. The question is where to place the package?
- in cob_robots: Bad because of cyclic dependency from/to cob_calibration_data
- as a standalone repository: will solve dependency structure, but will increase amount of overlays
- in cob_calibration_data: good, we just have to make that a multi-package-repo and rename the repository

Alternative would be to use the list out of cob_calibration_data. Pro: no cyclic dependencies because cob_calibration_data has no further dependencies and it doesn't hurt if all packages depend on cob_calibration_data

mathias-luedtke:
I vote for keeping the list `cob_calibration_data`, because it is easier to maintain.
A dependency on `xacro`does not hurt that much.

We can migrate to `cob_supported_robots` later if necessary.

> in cob_robots: Bad because of cyclic dependency from/to cob_calibration_data

this is not cyclic!
`cob_robot` -> `cob_supported_robots`
`cob_robot` -> `cob_hardware_config` ->  `cob_calibration_data` ->  `cob_supported_robots`

cyclic (with this PR):
`cob_hardware_config`  -> `cob_calibration_data` ->  `cob_hardware_config`

mathias-luedtke:
added `cob_supported_robots`

floweisshardt (approved):

mathias-luedtke:
@ipa-fmw: added your changes

fmessmer:
Once this is merged we need to update https://github.com/ipa320/cob_simulation/pull/130
Do we want a `cob_supported_environments` package? See https://github.com/ipa320/cob_environments/pull/116
Also, we want to make use of it in `cob_calibration_data`

fmessmer:
So, we are going to keep the `robotlist` in [cob_default_robot_behavior](https://github.com/ipa320/cob_robots/blob/indigo_dev/cob_default_robot_behavior/CMakeLists.txt#L8-L14) and [cob_moveit_config](https://github.com/ipa320/cob_robots/blob/indigo_dev/cob_moveit_config/CMakeLists.txt#L8-L17) because those are not implemented for all supported robots, correct?

cob_default_robot_config/package.xml
```diff
@@ -15,4 +15,6 @@
 
   <depend>roslaunch</depend>
 
+  <build_depend>cob_supported_robots</build_depend>
```
floweisshardt, line 18: why `build_depend`? other packages have `test_depend` defined

cob_hardware_config/package.xml
```diff
@@ -16,6 +16,8 @@
   <depend>roslaunch</depend>
   <depend>rostest</depend>
 
+  <build_depend>cob_supported_robots</build_depend>
```
floweisshardt, line 19: why `build_depend`? other packages have `test_depend` defined

cob_default_robot_config/package.xml
floweisshardt: due to using robotlist in install tags

cob_hardware_config/package.xml
floweisshardt: due to using robotlist in install tags

fmessmer:
`cob_calibration_data` is done in https://github.com/ipa320/cob_calibration_data/pull/115

PR #581, opened 2017-01-11
update moveit config for all robots

fmessmer:
This is just to not forget to do this now that the moveit_config structure has been updated...

 - re-run once new mesh for the cob4_torso (see https://github.com/ipa320/cob_manipulation/issues/89) is available
 - add cob4-5 as requested https://github.com/ipa320/cob_robots/pull/573#discussion_r93009286
 - other robots without an moveit_config: cob4-3, cob4-4, cob4-5, cob4-6, raw3-2, raw3-3, raw3-4, raw3-5, raw3-6
 - @ipa-bfb has an updated description + hardware_config for latest raw3-1 setup and an updated moveit_config for it

PR #582, opened 2017-01-11
[DoNotMerge] Export robotlist

fmessmer:
rebased version of #579

### DIFF
--- a/cob_bringup/CMakeLists.txt
+++ b/cob_bringup/CMakeLists.txt
@@ -1,13 +1,15 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(cob_bringup)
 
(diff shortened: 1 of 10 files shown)


fmessmer:
travis is likely to fail as not all moveit_configs are yet available....see #581

fmessmer:
```cob_hardware_config_ROBOTLIST``` could also be used in ```cob_default_robot_behavior```...

...but not all robots do have this implemented....what to do about it?
I think adding empty behaviors for the robots missing would allow us to add the behavior.launch to __all__ cob_bringup/robot.xml files keeping them as consistent as possible...but maybe there is a better solution with less copy-paste...

mathias-luedtke:
> I think adding empty behaviors for the robots missing

No, just keep a separate list for special features like moveit, behaviours etc.

PR #411, opened 2016-02-05
use handle_wait in pick behavior

fmessmer:
fixes https://github.com/ipa320/cob_robots/issues/387

needs to be tested on robot.

Issue #387, opened 2015-11-27
[ctitical] implement cob_default_robot_behaviour using handles


floweisshardt:
especially for torso movements and pick movements, we should check every command using the handles and abort if a previous movement cannot be completed as expected! 

```
handle = sss.move()
handle.wait()
if handle.get_error_code() != 0:
   ABORT
```
- define a handle for each `sss.move` command
- wait for movements to be finished before executing new movements
- do NOT use `spleep()` to synchronize movements (which can cause a collission)
- only continue if error_code is zero (==0)

This is currently missing for pick! ...disabling this button for now in the robot dashboard!

fmessmer:
fixed in #411

PR #591, opened 2017-01-19
Update behavior script to move the arms for cob4-7

NadiaHG:


ipa-fxm:
can you combine this with #411 and either pull them in or close?
related to https://github.com/ipa320/cob_robots/issues/387

fmessmer (changes requested):
can you combine this with #411 and either pull them in or close?
related to https://github.com/ipa320/cob_robots/issues/387

NadiaHG:
why?  they are two different things.... and different files. This pull request implements the **torso_front_cb behavior** for cob4-7, it is already tested on the robot and works. 

The https://github.com/ipa320/cob_robots/pull/411/files use handles for the **pick behavior** and it is still not tested.

fmessmer:
Well, IMHO it is related and I am trying to close old PRs and Issues...so if we test pick on cob4-7, we could also add it to cob4-2...it's all about keeping similar robots similar

floweisshardt (approved):
tested on cob4-7: OK

floweisshardt:
merging this one as #411 still has issues.

cob_default_robot_behavior/scripts/trigger_srvs_cob4-7.py
```diff
@@ -67,13 +67,28 @@
...
 def torso_front_cb(req):
-    sss.move_base_rel("base",[0,0,1.57],False)
-    sss.move("torso","front",True)
+    handle_arm_left = sss.move("arm_left","side", False)
+    handle_arm_right = sss.move("arm_right","side", False)
+    handle_arm_left.wait()
+    handle_arm_right.wait()
+    if handle_arm_left.get_error_code() == 0 and handle_arm_right.get_error_code() == 0:
+        sss.move_base_rel("base",[0,0,1.57],False)
+        sss.move("torso","front_down_full",True)
```
fmessmer, line 76: Well, then this condition is not safe, too, as we do not check whether `sss.move("torso", ...)` was correct....we simply `return TriggerResponse(True, "")`

PR #599, opened 2017-01-23
Enable collision monitor

fmessmer:
Now that torso mesh is fixed and collision matrix is updated, we could test the collision_monitor again...
Please test on robot before merging

### DIFF
--- a/cob_bringup/robots/cob4-2.xml
+++ b/cob_bringup/robots/cob4-2.xml
@@ -79,8 +79,7 @@
 		<include file="$(find cob_bringup)/tools/base_collision_observer.launch">
 			<arg name="robot" value="$(arg robot)"/>
 		</include>
-		<!-- <include file="$(find cob_bringup)/tools/collision_monitor.launch"> -->
-		<node pkg="rostopic" type="rostopic" args="pub /safety_controller/state_is_valid std_msgs/Bool 'data: true' -r10" name="fake_collission_monitor" output="screen"/>
+		<include file="$(find cob_bringup)/tools/collision_monitor.launch"/>
 
 		<include file="$(find cob_bringup)/tools/diagnostics_aggregator.launch">
 			<arg name="robot" value="$(arg robot)"/>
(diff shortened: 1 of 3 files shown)


fmessmer:
FIXME: closing xml tag is wrong...

fmessmer:
travis will be fixed after #602 is merged

fmessmer:
https://github.com/ipa320/cob_robots/pull/581 is merged...collision_monitor needs to be tested on hardware/simulation

fmessmer:
@ipa-nhg 
rebased and ready for testing on cob4-5

fmessmer:
I get an 
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Time is out of dual 32-bit range
[cob_collision_monitor_node-18] process has died [pid 28862, exit code -6, cmd /home/fxm/projects/indigo/care-o-bot/devel/lib/cob_collision_monitor/cob_collision_monitor_node /cob_collision_monitor_node/state_is_valid:=/safety_controller/state_is_valid __name:=cob_collision_monitor_node __log:=/home/fxm/.ros/log/eebacbb6-3bcd-11e7-86e7-0024bed884cf/cob_collision_monitor_node-18.log].
log file: /home/fxm/.ros/log/eebacbb6-3bcd-11e7-86e7-0024bed884cf/cob_collision_monitor_node-18*.log
```
when using it in simulation...investigating...

I think it's a race condition with `clock` and `/use_simtime` related to this lines https://github.com/ipa320/cob_manipulation/blob/indigo_dev/cob_collision_monitor/src/collision_monitor_node.cpp#L21-L23
as it works fine to run the launch file separately after gazebo is up...

@ipa-mdl FYI

fmessmer:
exception handling is added in https://github.com/ipa320/cob_manipulation/pull/114

NadiaHG:
as seen on cob4-5, the current meshes cause collisions even in acceptable configurations, e.g. within the cob4_hand (`gripper_right_finger_2_link` and `gripper_right_grasp_link`)....also the torso_center_link is using an enlarged collision mesh which causes collisions in certain default configuration, e.g. side...

thus, we would need to rework the meshes...

NadiaHG:
also there are configurations that lead to collision on hardware while geometric model is ok, i.e. `state_valid==true`

this is a matter of calibration!

head_shoulder_collision

ipa-rmb:
Still relevant or close?

PR #604, opened 2017-01-26
use xacro --inorder + move gazebo_ros_control plugin to component

fmessmer:
These PR-set allows to use latest [xacro](http://wiki.ros.org/action/fullsearch/xacro), i.e. e.g. conditional blocks

Also moves the per-component gazebo_ros_control plugin from the robot to the component. This is another step towards the robot-module-component structure we are tackling...

Related PRs are:
 - https://github.com/ipa320/cob_common/pull/208
 - https://github.com/ipa320/cob_manipulation/pull/97
 - https://github.com/ipa320/cob_control/pull/130
 - https://github.com/ipa320/schunk_robots/pull/72
 - https://github.com/ipa320/schunk_modular_robotics/pull/181
 - https://github.com/ipa320/cob_simulation/pull/136

### DIFF
--- a/cob_hardware_config/cob4-10/urdf/cob4-10.urdf.xacro
+++ b/cob_hardware_config/cob4-10/urdf/cob4-10.urdf.xacro
@@ -29,12 +29,6 @@
 
   <!-- composition of the robot -->
   <xacro:base name="base" use_old_joint_names="false"/>
-  <gazebo>
-    <plugin name="ros_control" filename="libhwi_switch_gazebo_ros_control.so">
-      <robotNamespace>base</robotNamespace>
-      <filterJointsParam>joint_names</filterJointsParam>
-    </plugin>
-  </gazebo>
 
   <xacro:torso name="torso" parent="base_link" dof1="false" dof2="true" dof3="true">
     <origin xyz="${torso_x} ${torso_y} ${torso_z}" rpy="${torso_roll} ${torso_pitch} ${torso_yaw}" />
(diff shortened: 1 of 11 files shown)


mathias-luedtke:
What is the matter with the last commit?
I prefer lowercase true/false.

fmessmer:
me, too...but it's because of lines like [this one](https://github.com/ipa320/cob_common/pull/208/files#diff-7a500bdef900893334795e9776e14b01R145) where I use [conditional blocks](http://wiki.ros.org/action/fullsearch/xacro#Conditional_Blocks) that need to be pure python....i.e. False/True...it's a bummer

mathias-luedtke:
The documentation says: "The expression needs to evaluate to "0", "1", "true", or "false", otherwise an error will be thrown.", so this should still work somehow.
If not, it is a regression bug, which needs to be reported.
It might have even be fixed, but wasn't backported to indigo..

Simple fix:
```
<xacro:if value="${active}=='true'">
```
or
```
<xacro:if value="'${active}'">
```

fmessmer:
I opened an issue: https://github.com/ros/xacro/issues/145
Let's see what the comments are...
Otherwise, I will test one of @ipa-mdl's proposed workarounds...

fmessmer:
Affects https://github.com/ipa320/cob_common/pull/208 and this PR + https://github.com/ipa320/schunk_modular_robotics/pull/181 which already has been merged

fmessmer:
https://github.com/ipa320/cob_common/pull/208 is updated using @ipa-mdl's workaround:
 > `<xacro:if value="${(dof1 == 'true') or (dof2 == 'true') or (dof3 == 'true')}">`

Thus, `True`/`False` is reverted in this PR and for https://github.com/ipa320/schunk_modular_robotics/pull/181 `True`/`False` is reverted in https://github.com/ipa320/schunk_modular_robotics/pull/182

PR #607, opened 2017-01-30
removed needless image_proc node

benmaidel:
usb_cam publish rectified image by itselve

### DIFF
--- a/cob_bringup/drivers/usb_camera_node.launch
+++ b/cob_bringup/drivers/usb_camera_node.launch
@@ -14,6 +14,4 @@
 		<param name="camera_name" value="$(arg camera_name)"/>
 		<param name="camera_info_url" value="package://cob_calibration_data/$(arg robot)/calibration/cameras/$(arg camera_name).yaml"/>
 	</node>
-	<node name="image_proc" pkg="image_proc" type="image_proc" ns="$(arg camera_name)"/>
-
 </launch>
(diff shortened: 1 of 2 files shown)


fmessmer:
Could you remove the `image_proc` from the `cob_controller_configurations_gazebo` launch files, where they are used for `head_cam`, e.g. [cob4-2](https://github.com/ipa320/cob_robots/blob/indigo_dev/cob_controller_configuration_gazebo/launch/robots/default_controllers_cob4-2.launch#L20)...
In the end, it might be, that `image_proc` is nowhere else used within cob_robots...then we can remove the dependency, too

see https://github.com/ipa320/cob_robots/search?utf8=%E2%9C%93&q=image_proc

benmaidel:
Don't we need `image_proc` in simulation? Just the `usb_cam` driver is already publishing rectified images if a calibration file is provided to the driver. Is gazebo publishing rectified images as well?

fmessmer:
Öhm, don't know? How can we verify this?
Maybe there is a setting in the gazebo camera plugin that we use for the USB cam

floweisshardt:
are you sure that the usb driver is publishing a rectified image? http://wiki.ros.org/usb_cam does not say anything about that.

benmaidel:
If you start the usb_cam driver there is a topic named `image_rect` which is also published. `usb_cam` is using image_transport for publishing images. Now i am not sure if node is really publishing rectified image or if it is a bug inside image_transport...

benmaidel:
Can someone check the topics published by usb_cam under indigo? I'm using kinetic

benmaidel:
`image_transport` under kinetic is publishing a rectified image. Just checked with my notebooks webcam!

benmaidel:
`image_transport` under `indigo` isn't publishing rectified images. So this PR should NOT be merged! Instead we need to reopen it to kinetic_dev!

PR #608, opened 2017-01-31
[kinetic] removed unused image_proc node

benmaidel:
On kinetic image_pipeline is already publishing rectified images. So no need to start extra node for rectification.

See this PR https://github.com/ipa320/cob_robots/pull/607


benmaidel:
checked with @ipa-fxm. image_proc is still needed for rectified images in `indigo` and `kinetic`.

PR #609, opened 2017-02-01
Raw3 5 config

andreeatulbure:
Working configs for raw3-5

We changed canopen_base.launch argument values to default because we needed to set value.


cob_bringup/components/canopen_base.launch
```diff
@@ -1,13 +1,13 @@
 <?xml version="1.0"?>
 <launch>
 	<arg name="robot"/>
-	<arg name="component_name" value="base"/>
+	<arg name="component_name" default="base"/>
```
fmessmer: I guess `value` vs. `default` was intended by @ipa-mdl, as it is a base-specific launch file...

Don't know about the parameters below

cob_bringup/robots/raw3-5.xml
```diff
@@ -39,29 +39,34 @@
...
+			<arg name="robot" value="$(arg robot)" />
 		</include>
-		<include file="$(find cob_bringup)/tools/emergency_stop_monitor.launch">
-			<arg name="robot" value="$(arg robot)"/>
+		<include file="$(find cob_bringup)/drivers/battery_voltage.launch" >
+			<arg name="robot" value="$(arg robot)" />
 		</include>
-		<include file="$(find cob_bringup)/drivers/battery_voltage.launch">
-			<arg name="robot" value="$(arg robot)"/>
+		<!-- problems when using battery_filter
```
ipa-fxm: What's the problem here?
Is it worth an issue?

fmessmer, line 49: What's the problem here?
Is it worth an issue?

fmessmer (commented):
Other than that, @ipa-mdl can best approve this PR

cob_bringup/components/canopen_base.launch
mathias-luedtke: These should be values, I just kept the argument for having the config at the top and not spreaded.

ipa-mdl: These should be values, I just kept the argument for having the config at the top and not spreaded.

cob_bringup/robots/raw3-5.xml
```diff
@@ -39,29 +39,34 @@
...
-			<arg name="robot" value="$(arg robot)"/>
-		</include>-->
-		<include file="$(find cob_bringup)/tools/diagnostics_aggregator.launch">
-			<arg name="robot" value="$(arg robot)"/>
+		<include file="$(find cob_bringup)/components/canopen_base.launch" >
+			<arg name="robot" value="$(arg robot)" />
+			<arg name="component_name" value="base" />
+			<arg name="can_device" value="can1" />
+			<arg name="interval_ms" value="20" />
+			<arg name="use_external_sync" value="true" />
```
mathias-luedtke: external sync was not tested properly for use with all components.
It will get the default somewhen (in `canopen_base.launch`)

cob_hardware_config/raw3-5/config/base_controller.yaml
```diff
@@ -20,17 +21,21 @@ twist_controller:
       spring: 15.0 #10.0
       damp: 2.75 #2.5
       virt_mass: 0.1 #0.1
-      d_phi_max: 12.0 #100.0
+      d_phi_max: 10.0 #100.0
       dd_phi_max: 40.0 #100.0
   wheels: &wheels
     - steer: fl_caster_rotation_joint
       drive: fl_caster_r_wheel_joint
+      steer_neutral_position: -180.0
```
mathias-luedtke, line 29: Why do you need to set -180 for three wheels?

cob_bringup/robots/raw3-5.xml
andreeatulbure: when battery_voltage_filter.launch is used, the raw3-5 did not initialize and did not move

andreeatulbure: when battery_voltage_filter.launch is used, the raw3-5 did not initialize and did not move

cob_hardware_config/raw3-5/config/base_controller.yaml
andreeatulbure: We set -180 to three wheels because they were driving in the wrong direction after initialization. Only one wheel was driving in the correct direction

mathias-luedtke: > Only one wheel was driving in the correct direction

The neutral postion position should not affect driving.
It is just a default target for the steer wheels after start-up.

andreeatulbure: When the neutral position was set to 0 degrees the 3 wheels did move back when the command was moving forward. Changing the neutral position to -180 degrees, the wheels moved 180 degrees and  moved forward when the command from the joystick was to move forward. If it does not depend on the neutral position, should we change the encoder values to get the movement in the correct direction?

mathias-luedtke: The controller does not use the neutral position in any way for steering control.
I guess the encoder parameters aren't set-up properly, especially the sign.

andreeatulbure: Ok. Thanks for the information.

mgruhler: note that you can also define what the elmo considers to be positive/negative turning direction in the controller itself. @ipa-mdl is this configured somewhere in those files? If not, this could be the issue here that this has been flashed to the elmo.

mathias-luedtke: @ipa-mig: there are plenty of options, I think the easiest is to prepend a sign to the (velocity) factors. 
(The old driver used https://github.com/ipa320/cob_robots/blob/indigo_dev/cob_hardware_config/raw3-5/config/base/CanCtrl.ini#L150)
In the end the direction of the motion should match the URDF.

cob_hardware_config/raw3-5/config/base_controller.yaml
```diff
@@ -24,15 +24,15 @@ twist_controller:
...
-    - steer: fl_caster_rotation_joint
-      drive: fl_caster_r_wheel_joint
-      steer_neutral_position: -180.0
     - steer: bl_caster_rotation_joint
       drive: bl_caster_r_wheel_joint
       steer_neutral_position: -180.0
     - steer: br_caster_rotation_joint
       drive: br_caster_r_wheel_joint
       steer_neutral_position: -180.0
+    - steer: fl_caster_rotation_joint
```
mathias-luedtke: If I remember it correctly, the order was important and must be counter-clockwise.

andreeatulbure: In our case, the order in this file did not change anything. The order in the base_driver.yaml file influenced our initialization

mathias-luedtke: >  The order in the base_driver.yaml file influenced our initialization

Yeah, that's fine. 

> In our case, the order in this file did not change anything.

It should influence the odometry formulas, but I am not sure.

fmessmer:
So, is this PR good to merge as is...or do you need to verify the order of the joint names (or anything else) again?

@andreeatulbure @destogl 
Could I get a last confirming comment from your side?

mathias-luedtke:
IMHO the joints must be listed as requested, otherwise we cannot ensure that everything works as expected.

fmessmer:
@destogl @andreeatulbure 
Please consider @ipa-mdl's request and test on hardware.
Then, finally squash this PR appropriately.

We have PR #622 pending which will break this PR...and I want to merge #622 this week!

destogl:
@andreeatulbure can you do this this week?

destogl:
@ipa-fxm: as we will not be able to merge this today you are free to merge #622 and break this PR. In the next weeks we will update this PR accordingly...

Thanks for your patience and sorry if we slow down your development.

andreeatulbure:
We tried to squash as many commits as possible, but we have several merges in the repository and therefore it is hard to rebase. I worked about 2 hours trying to squash more commits and I would say that it takes too much time to do it. Can you merge it with the remaining commits?

fmessmer:
@andreeatulbure @destogl 
LGTM from my side...Thanks for your effort.

@ipa-mdl 
One last review from your side?
In case you approve, I'll merge...

fmessmer:
@ipa-mdl 
👍 ?

cob_hardware_config/raw3-5/config/diagnostics_analyzers.yaml
```diff
@@ -38,7 +38,7 @@ analyzers:
           base:
```
mathias-luedtke, line 38: cob4 config:
```
      base:
        type: diagnostic_aggregator/GenericAnalyzer
        path: Base
        contains: 'base/'
```

mathias-luedtke (approved):
If it works..

fmessmer (approved):

fmessmer:
I'll squash-merge this now.
We will come across this anyway as we need to update #622 (sorry @ipa-mdl, but I guess this way it's easier...)

andreeatulbure:
we noticed that in some situations the roboter shakes while being moved, therefore we changed some parameters in the base_controller.yaml file

fmessmer:
@andreeatulbure feel free to open a new PR with the updated config...
this PR did it's purpose which was to swap from old base driver to ros_canopen...

PR #614, opened 2017-02-08
fixed cam3d_down camera calibration

benmaidel:
camera default link coordinate system:
x = forward
y = left
z = up

cameras optical frame coordinate system:
z = forward
x = right
y = down

realsense driver is publishing tf transform for base and optical frames so we also have to stick to the defaults.

See [REP](http://www.ros.org/reps/rep-0103.html)


benmaidel:
tested with @ipa-fmw

floweisshardt (approved):
tested on cob4-2: OK

fmessmer:
do this for cob4-5/cob4-7/cob4-10, too?

fmessmer:
done in https://github.com/ipa320/cob_robots/pull/616

PR #616, opened 2017-02-10
fixed camera down camera calibration for all robots

fmessmer:
@ipa-bnm 
applies https://github.com/ipa320/cob_robots/pull/614 to all robots

### DIFF
--- a/cob_hardware_config/cob4-10/urdf/calibration_default.urdf.xacro
+++ b/cob_hardware_config/cob4-10/urdf/calibration_default.urdf.xacro
@@ -80,9 +80,9 @@
 	<xacro:property name="def_torso_cam3d_down_x" value="0.15908"/>
 	<xacro:property name="def_torso_cam3d_down_y" value="0.0"/>
 	<xacro:property name="def_torso_cam3d_down_z" value="0.14005"/>
-	<xacro:property name="def_torso_cam3d_down_roll" value="2.18166"/> <!--2.18166-->
-	<xacro:property name="def_torso_cam3d_down_pitch" value="0.0"/>
-	<xacro:property name="def_torso_cam3d_down_yaw" value="1.57079"/> <!--1.57079-->
+	<xacro:property name="def_torso_cam3d_down_roll" value="0.0"/>
+	<xacro:property name="def_torso_cam3d_down_pitch" value="0.61086"/>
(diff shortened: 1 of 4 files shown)


benmaidel (approved):

floweisshardt (approved):

NadiaHG:
did you check if the point cloud vs tf are also correct on simulation?

fmessmer:
@ipa-nhg, no, I didn't test in simulation, I was just applying @ipa-bnm's PR to all other robots...
I guess we can consider this when working on https://github.com/ipa320/cob_common/pull/206. For now, it seems to be correct with Hardware...but it might never have been correct in simulation...so nothing breaks here!

NadiaHG:
realsense is a special case, If you look the urdf file of the realsense camera https://github.com/ipa320/cob_common/blob/indigo_dev/cob_description/urdf/sensors/realsense.urdf.xacro we are not defining the optical frames, the realsense_driver is the responsible of the transformations https://github.com/intel-ros/realsense/blob/indigo-devel/realsense_camera/launch/r200_nodelet_rgbd.launch#L40 , but on simulation we are using a standard gazebo plugin. Theoretically  the x-axis for the camera_link should point to the "view direction", but I am not sure if realsense is following also this standard. 

Maybe we should set the parameter` publish_tf` to `false` and define the frames in the URDF file to make if working on simulation (?).

benmaidel:
>Theoretically the x-axis for the camera_link should point to the "view direction", but I am not sure if realsense is following also this standard.

realsense is following the [Standard](http://www.ros.org/reps/rep-0103.html). The x axis of camera_link should point into view direction and the z axis of the optical frame should point into view direction as well. See my first comment here https://github.com/ipa320/cob_robots/pull/614

>Maybe we should set the parameter `publish_tf` to false and define the frames in the URDF file to make if working on simulation (?).

We should not disable `publish_tf` and add the frames to urdf. The realsense driver reads the camera calibration from the hardware and creates `static_transform_broadcaster` for these calibration values.

Perhaps we could create one urdf for simulation and one for the real hardware?

NadiaHG:
Realsense is not working on simulation, neither with the old version. This pull request can be merged and I will open a new issue for the simulation.

NadiaHG (approved):

fmessmer:
related issue is https://github.com/ipa320/cob_common/issues/212

PR #623, opened 2017-02-17
change u_max to meet the measured values

jabbenseth:


NadiaHG (approved):

benmaidel:
I think we have to recheck this! Or is it really possible that the Battery reaches a maximum value of 100 Voltage?

jabbenseth:
No. The name is quite misleading. The only thing this parameter does is modifying the factor of the voltage divider. Like analog_read/(phidgets_max - phidgets_min) *voltage_max. During testing I found the phidgets board publishes a value of around 510 if the voltage is 51.0v and phidgets_max is set to 999 and phidgets_min to 0. What we basically need is to divide the analog reading by 10 to have the correct output reading. So in fact 99.9 would be the more precise value. Renaming the parameter to voltage_divider_factor would be more understandable

fmessmer:
Is this PR related https://github.com/ipa320/cob_driver/pull/298
Could you test/finalize #623 and https://github.com/ipa320/cob_driver/pull/298 together?

benmaidel:
Ok, right. The Parameter is really misleading! I would vote to rename the parameter to `voltage_divider_factor` as @ipa-jba has mentioned.

jabbenseth:
@ipa-fxm in my opinion they are independent. however they fix (#623) /introduce features(https://github.com/ipa320/cob_driver/issues/298) on the same node

jabbenseth:
strongly coupled with https://github.com/ipa320/cob_driver/issues/330

floweisshardt (approved):

fmessmer (approved):

PR #624, opened 2017-02-21
changes for using fake docking and power usage

souravran:


cob_bringup/tools/docker_control.launch
```diff
@@ -32,4 +33,8 @@
 	<node pkg="cob_docker_control" type="docker_control.py" respawn="false" name="docker_control" output="screen" unless="$(arg multi_station)">
 	</node>
 
+	<!-- virtual charging and power usage for simulation -->
+	<node if="$(arg sim)" pkg="cob_docker_control" type="fake_docking.py" respawn="false" name="fake_docker_control" output="screen" >
+	</node>
```
fmessmer: can we use single line notation?
`<node if="$(arg sim)" pkg="cob_docker_control" type="fake_docking.py" name="fake_docker_control" respawn="false" output="screen" />`

cob_bringup/tools/docker_control.launch
```diff
@@ -32,4 +33,8 @@
 	<node pkg="cob_docker_control" type="docker_control.py" respawn="false" name="docker_control" output="screen" unless="$(arg multi_station)">
 	</node>
```
fmessmer: This node does not exist anymore...what happened to it?

cob_bringup/tools/docker_control.launch
fmessmer: Not sure, but does any of the other nodes than need a `unless="$(arg sim"` tag?

fmessmer (changes requested):

cob_bringup/tools/docker_control.launch
souravran: I had the same question but my best guess is that the node may be obsolete or has been updated to ``multi_station_docker_control.py`` and was not removed from here.

cob_bringup/tools/docker_control.launch
souravran: Even I am not sure about the ``unless="$(arg sim)"`` tag, simulation works fine though.

cob_bringup/tools/docker_control.launch
fmessmer: Do you see any error message about "Node not there"?

Please remove the node from the launch file

souravran: No I dont get any error. Alright.

fmessmer: There still is a `docker_control.py` in https://github.com/ipa320/cob_substitute/tree/indigo_dev/cob_docker_control/scripts, maybe it uses this one...
Do you have an overlay of `cob_substitute`? Or do you use a released version?
Can you put one in your workspace and remove the `docker_control.py`? Then you should get an error....
Please provide a PR for cob_substitute where the script is removed...

In the same way, we might need to add a `fake_docking.py` to `cob_substitute`!

souravran: No I dont have cob_substitute, will open an PR after testing with cob_substitute overlay.

fmessmer:
requires https://github.com/ipa320/cob_driver/pull/321 for travis to be happy

@ipa-fmw 
Do we need to `unless=$(arg sim)` the `multi_station_docker_control.py`?
(What's the purpose of this file anyway?)

@souravran 
PR to `cob_substitue` still missing

floweisshardt:
we typically use multi_docking, which can dock e.g. to the charging station and the travel box. This works too for simulation (together with fake_docking providing correct power_state), so no need for `arg sim` in multi_docking.
But we still need `arg sim` for fake_docking

fmessmer (approved):
cob_substitute has been updated: https://github.com/ipa320/cob_substitute/pull/43

PR #634, opened 2017-03-14
Setup cob4 paul stuttgart

NadiaHG:
- paul configuration for cob4-7 
 - remove cob4-10 robot,not needed (this was cob4-7 with arms)

PR #635, opened 2017-03-16
Cob4 paul stuttgart

fmessmer:
#634 rebase, sqashed and sorted


mathias-luedtke:
files are identical to #634 (meld compare)

PR #638, opened 2017-03-20
skip install space tests

mathias-luedtke:


mathias-luedtke:
There is something wrong with cob_extern.
Perhaps the ExternalProject solution does not cover all cases.

fmessmer:
I saw it...I'm investigating...

mathias-luedtke:
I have tested it locally with `cob_phidgets` and `libphidgets` only, and it worked..

fmessmer:
I also don't see what the problem is...all `libphidgets` headers and libs are where they are supposed to be - for both devel and install....

strange is also, that only kinetic is complaining...
(jade complains because of something different #639)

fmessmer:
let's see how it goes now...

mathias-luedtke:
:tada:

mathias-luedtke:
@ipa-fxm: 
This option can be added safely. Per default the full workspace gets cleaned and rebuilt to install space just to run all installed *.test files.
Currently this means that the install rules do not get checked, but this will change after https://github.com/ros-industrial/industrial_ci/pull/150 was merged

fmessmer (approved):

fmessmer (approved):

PR #639, opened 2017-03-20
update .rosinstall file

fmessmer:


fmessmer:
well, I wanted to see jade fixed as well...but it's getting too complicated:
 - `realsense_camera` is not released for jade
 - and `realsense` from source is broken...

PR #654, opened 2017-04-12
Separate sensor actor (1)

fmessmer:
related to #437 
 > extract sick scanners from base urdf + then remove nav configs for mlr bases as they do not have laser scanners anymore

merge together with
 - https://github.com/ipa320/cob_common/pull/217
 - https://github.com/ipa320/cob_navigation/pull/86

@ipa-nhg please review

Issue #422, opened 2016-02-12
switch raws and cob3s to new canopen base driver


floweisshardt:
currently done and tested for cob4-2 (see #405).
Instructions see https://github.com/ipa320/cob_robots/pull/405#issuecomment-182945753

@destogl will test with raw3-5

mathias-luedtke:
It does not work for cob-3, because steer and drive are coupled mechanically.
If we want to support them, first https://github.com/ros-industrial/ros_canopen/issues/101 needs to be addressed and second a specialised motor node needs to be written.

destogl:
@ipa-fmw: Thanks for nice instructions.

I will do it. But probably not before march because currently I have other student working on the robot using powercube but I will try to port also the powercube to use SocketCAN (it shouldn't be too hard) and than I test every thing.

mathias-luedtke:
@destogl: a SocketCAN-based powercube driver sound interesting. `socketcan_interface` already encapsulates most aspect. In addition there is a new topic-base interface (http://wiki.ros.org/socketcan_bridge)

destogl:
@ipa-mdl Thanks! I prefer library-like interface for this kind of stuff. We should have already one implementation which is independent of socketcan_interface, I am looking to integrate it in the next days.

off-topic: There are also few other nice stuff we did with powercube (https://github.com/ipa320/schunk_modular_robotics/pull/156) but it is too dirty for merge. I will clean it an make few PRs.

fmessmer:
Take care and note of https://github.com/ipa320/cob_control/issues/98!

Old ELMOs need to be set to `UM=2` persistently and velocity sensor needs to be enabled (see #445)

andreeatulbure:
Hello,

I and @destogl tried to test the ros_canopen config #405 on the raw3-5. We followed the config steps described in #405 
System set-up:
- overlay cob_driver with ipa320/cob_driver#258
- source overlay https://github.com/ros-industrial/ros_canopen.git
- source overlay: https://github.com/ros-controls/ros_control.git (maybe optional)
- disable peak driver and activate socketcan peak_usb
- configure socketcan with bitrate 1000000

Config (based on #417) :
-  in base_driver.yaml:
        add missing wheel
        update homing_offset (EncOffsetIncr in base/CanCtrl.ini)
        check homing_event (HomingDigIn in base/CanCtrl.ini)
        add "606A": "1" # read velocity from velocity sensor
- in base_controller.yaml:
[...]

PR #655, opened 2017-04-14
[WIP] restructure hardware config

fmessmer:
related to #437 
includes #654 but will be rebased properly once that is merged...

trying to find a concept for this...this is rather a testing PR atm!


fmessmer:
in https://github.com/ipa320/cob_robots/pull/655/commits/2f0866b2ce5c0c38eabce1d74755128a4ba16115 I introduced launch files for each of our base modules (naming according to Verbautliste)...

As a first example, I then extracted the configuration for the laser scanners (sick_s300, sick_lms1xx) from the robot-specific configuration (i.e. `cob_hardware_config/robots`) and split it into component-specific yaml (i.e. driver settings) and module-specific settings (e.g. host/port, scan_interval).

I think, this is pretty much what we discussed...
Step by step we could now move more and more module-/components-specific configurations from the robot-specific `cob_hardware_config/robots` and create according module launch files and component yamls...

@ipa-nhg @ipa-fmw @ipa-mig @ipa-mdl  FYI

fmessmer:
I moved on a bit further...restructuring the actuators...I'm not sure whether arms sensorrings and grippers should actually be `modules` themselves....


Latest commit https://github.com/ipa320/cob_robots/pull/655/commits/7143c9c2a091a70036e0ceb0c268dfb708779373 is `roslaunch-dump`-verified
Feedback wanted!

cob_bringup/bundles/base_cob4.launch
```diff
@@ -1,16 +1,15 @@
...
-	<arg name="robot"/>
-
+	<arg name="base"/>
 	<arg name="can_device" default="can0"/>
 	<arg name="pkg_hardware_config" default="$(find cob_hardware_config)"/>
 	<arg name="sim" default="false"/>
 
-	<arg name="driver_yaml" value="$(arg pkg_hardware_config)/robots/common/cob4_base_driver.yaml"/>
-	<arg name="controller_yaml" value="$(arg pkg_hardware_config)/robots/common/cob4_base_controller.yaml"/>
+	<arg name="driver_yaml" value="$(arg pkg_hardware_config)/modules/base/common/base_cob4_driver.yaml"/>
```
floweisshardt, line 8: I'd prefer to keep `cob4` prefix before `base` prefix. what's your rational to change this?

cob_bringup/bundles/schunk_powercube_legacy.launch
```diff
@@ -1,23 +1,27 @@
 <?xml version="1.0"?>
```
floweisshardt, line 1: do we still have supported robots using `legacy` version of schunk components? if not, we can get rid of this.

cob_bringup/robots/base_solo.launch
```diff
@@ -16,11 +17,12 @@
...
 	<node pkg="robot_state_publisher" type="state_publisher" name="robot_state_publisher"/>
 
-	<include unless="$(arg use_old_base_drive_chain)" file="$(find cob_bringup)/components/cob4_base.launch">
+	<include unless="$(arg use_old_base_drive_chain)" file="$(find cob_bringup)/bundles/base_cob4.launch"/>
+		<arg name="base" value="$(arg base)"/>
 	</include>
 
-	<include if="$(arg use_old_base_drive_chain)" file="$(find cob_bringup)/components/legacy_base.launch">
-		<arg name="robot" value="$(arg robot)"/>
+	<include if="$(arg use_old_base_drive_chain)" file="$(find cob_bringup)/bundles/base_legacy.launch">
```
floweisshardt, line 24: do we still have supported robots with `legacy` version of base?

floweisshardt (commented):
looks good in general, although I need more explanation about the differences between component, module and bundle...

cob_bringup/robots/base_solo.launch
mgruhler: Afaik, the r@w's could be switched to the canopen version, but not the cob3-X Bruno just reactivated...

cob_bringup/bundles/base_cob4.launch
fmessmer: as we are grouping by module, I'd have the "module" as prefix...grouping the launch files, too

cob_bringup/bundles/schunk_powercube_legacy.launch
fmessmer: raw3-1 torso is using the legacy schunk_powercube_chain

cob_bringup/robots/base_solo.launch
fmessmer: currently it's us not moving on here....the only robots using legacy base are `raw3-1` and `raw3-3` - `raw3-5` could be used as a template..see #422

cob_hardware_config/components/sick_s300/sick_s300.yaml
```diff
@@ -1,9 +1,4 @@
-port: /dev/ttyUSB0
 baud: 500000
 scan_duration: 0.025 #no info about that in SICK-docu, but 0.025 is believable and looks good in rviz
 scan_cycle_time: 0.040 #SICK-docu says S300 scans every 40ms
-inverted: false
-scan_id: 7
-frame_id: base_laser_rear_link
 publish_frequency: 12
-scan_intervals: [[-2.0, 2.0]] #[rad] these intervals are included to the scan
```
mgruhler, line 9: I'm still not sure if it makes sense to have those parameters only "component-specific". The ones you pull out are definitely depending on the real instance of the sensor. The rest could, however, in theory be changed as well.

What I don't like too much is that we now have the parameters in two places. The yaml file here and the launch args in the respective modules. This probably makes sense for many nodes, but especially for this one, one could argue that if those cannot/should not be changed depending on the real instance, they could/should be hard-coded and not parameters at all.

I fully understand the idea behind this, but I think there are cases where it is difficult to decide what is depending on the real sensor and what on the sensor type...

cob_hardware_config/components/sick_lms1xx/sick_lms1xx.yaml
```diff
@@ -0,0 +1,3 @@
+angle_resolution: 0.25 #deg
+scan_frequency: 25 #hz
+set_config: false
```
mgruhler, line 3: Here, the same holds as for the s300 below. Those parameter might need to be changed depending on the instance of the sensor, not the component of "lms1xx" in general.

You can configure the scanner to have a resolution of `0.25` or `0.5` degrees (probably even some more). You could even have one of each on one single robot.

Do you see my point?

NadiaHG:
My idea was a little bit different, a module is for example the cob4-t7 and this module contains the torso axis,  3 pcs,  light, 3 cameras.. 
And the robots/cob4-X should be a combination of modules, in case that we exchange the modules this combination could be auto-generated, in an ideal world only the urdf and the launch file...  but this will be a  next step...

cob_hardware_config/components/sick_s300/sick_s300.yaml
fmessmer: It's not even just that, i.e. deciding whether it's "instance-general" (identical for all instances of the same component/module/robot) or "instance-specific" (possibly different for each instance of the same component/module/robot)...

...I had severe brain-knots when needing to decide whether to call s.th. an instance, a component, a module, a bundle....or simply keep it in robot...

Thus, so far, I don't see the real benefit of this restructuring myself...I know where we want to go, but I have no concept for it (yet)...
However, I think we should go for an approach where we combine/generalize things that are **currently** identical - even if at a future point we might introduce a new instance that requires different configuration....I think we should handle such case once it occurs - rather than already considering the most general, galactical configuration structure possible...it's all about making our life easier **atm**, i.e. **now**, i.e. with the requirements we **currently** have (not the ones we might have in the future - maybe)

If someone knows a good concept to do so, please provide input @ipa-nhg @ipa-mdl

fmessmer: Apart from that, I once heard about a "templatable" yaml approach (https://github.com/ros-industrial/universal_robot/issues/144), but I never got an answer about how to use it...
Does someone know more about it?

fmessmer:
@ipa-nhg 
Do you want to propose your own concept yourself?
Should we do this in a (e.g. half-day) hackathon-style session? maybe next week?
(@ipa-fmw @ipa-mig feel free to join if you like...)

floweisshardt:
yes, I'd like to join. let's plan on monday during our sprint planning session

fmessmer:
closing (unmerged) due to https://github.com/ipa320/cob_robots/issues/437#issuecomment-299639257
I'll keep the branch though...

PR #658, opened 2017-04-20
Added a launch file to display a urdf.xacro model

NadiaHG:


cob_hardware_config/CMakeLists.txt
```diff
@@ -25,3 +25,7 @@ install(DIRECTORY robots test
 install(FILES upload_robot.launch
   DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}
 )
+
+install(FILES xacrodisplay.launch
```
fmessmer: better name `display_robot.launch`?
(you could also combine the two `FILES` install tags...

cob_hardware_config/xacrodisplay.launch
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0"?>
```
fmessmer: `upload_robot.launch` uses tabs for indentation

cob_hardware_config/xacrodisplay.launch
```diff
@@ -0,0 +1,23 @@
...
+<launch>
+  <arg name="robot" default="$(optenv ROBOT !!NO_ROBOT_SET!!)"/>
+  <arg name="gui" default="false" />
+  <arg name="rvizconfig" default="$(find urdf_tutorial)/urdf.rviz" />
```
fmessmer: these introduces an `exec_depend>` to `urdf_tutorial`...
how about copying that file...or simply use `cob_hardware_config/robots/$(arg robot)/$(arg robot).rviz` (should be fine enough having a RobotModel loaded to the displays...

cob_hardware_config/xacrodisplay.launch
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0"?>
+<launch>
+  <arg name="robot" default="$(optenv ROBOT !!NO_ROBOT_SET!!)"/>
+  <arg name="gui" default="false" />
```
fmessmer: default `true`?

fmessmer (commented):
some comments

cob_hardware_config/xacrodisplay.launch
NadiaHG: the `cob_hardware_config/robots/$(arg robot)/$(arg robot).rviz` includes the cameras, the sensors data, map... all of this is not needed for the purpose of the file. I propose copy the file to `cob_hardware_config/robots/common/display_robot.rviz`, is this ok?

cob_hardware_config/xacrodisplay.launch
NadiaHG: I prefer false, I only use the gui modus to check the limits...

cob_hardware_config/display_robot.launch
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0"?>
+<launch>
+
+	<arg name="robot" default="$(optenv ROBOT !!NO_ROBOT_SET!!)"/>
+	<arg name="gui" default="false" />
+	<arg name="rvizconfig" default="$(find cob_hardware_config)/robots/config/display_robot.rviz" />
```
fmessmer: the path of the file is `robots/common`!

fmessmer (changes requested):

cob_hardware_config/xacrodisplay.launch
floweisshardt: I vote for using the existing rviz configs. it does not hurt if there are laser scanners and cameras defined in the config. if we introduce a new rviz config thats one more file we have to maintain without gaining mayor benefits.

floweisshardt (changes requested):
I vote for using the existing rviz configs. it does not hurt if there are laser scanners and cameras defined in the config. if we introduce a new rviz config thats one more file we have to maintain without gaining mayor benefits.

fmessmer (approved):

fmessmer:
I leave it to @ipa-fmw and @ipa-nhg to decide about the config and then merge this PR...I'm ok with both options...

floweisshardt:
@ipa-nhg: let's fight :-)

floweisshardt (approved):

PR #659, opened 2017-04-21
use topic_status_monitor (0.5)

fmessmer:
proposes to use the diagnostic_updater-based node from https://github.com/ipa320/cob_command_tools/pull/170 as a drop-in replacement for `hz_monitor.py`
no need to change/modify any yaml (hz_monitor.yaml, diagnostics_aggregator.yaml)...all works well...it just provides more information about topic status see https://cloud.githubusercontent.com/assets/508006/22161614/7f14c6c4-df4b-11e6-933f-78afe2c41727.png

@ipa-fmw @ipa-nhg

### DIFF
--- a/cob_bringup/tools/hz_monitor.launch
+++ b/cob_bringup/tools/hz_monitor.launch
@@ -6,9 +6,8 @@
 	<arg name="pkg_hardware_config" default="$(find cob_hardware_config)"/>
 
 	<!-- start hz monitor node -->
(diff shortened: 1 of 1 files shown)


floweisshardt:
please remove the "old" ``hz_monitor.py`, so that we do not have two versions to maintain.

Before merging we should test on cob4-paul-stuttgart

fmessmer:
### ToDo:
 - update respective yamls according to https://github.com/ipa320/msh/issues/708
 - remove old `hz_monitor.py`

fmessmer:
included in #664 and ipa320/msh#711

PR #660, opened 2017-04-26
[cob4-5] Feature/enable head 

NadiaHG:


NadiaHG:
Hardware configuration tested on cob4-5 for sensorring and head:
- Init :+1: 
- Check joint_states :+1: 
- Move using rqt :+1: 
- Recover after emergency stop :+1: 

:warning: the command_gui positions have to be tested 

@ipa-fmw @ipa-fxm this branch should also works on cob4-7

cob_hardware_config/cob4-5/config/head_driver.yaml
```diff
@@ -0,0 +1,46 @@
+bus:
```
NadiaHG, line 1: theoretically we should remove these lines because we are already setting  the values on the launch file https://github.com/ipa320/cob_robots/blob/feature/enable_head/cob_bringup/drivers/canopen_402.launch#L21 , there parameters are not being used/overwritten , see https://github.com/ipa320/cob4/issues/349#issuecomment-296593818

fmessmer:
will be finalized in #662

PR #666, opened 2017-05-11
Update cob4-5 setup

NadiaHG:


cob_bringup/components/cam3d_sick_visionary.launch
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0"?>
```
floweisshardt: we should call this file `..._visionary_t.launch`, same as in drivers

cob_bringup/components/cam3d_sick_visionary.launch
```diff
@@ -0,0 +1,17 @@
...
+<launch>
+	<arg name="robot"/>
+	<arg name="name"/>
+	<arg name="sim" default="false"/>
+
+	<include file="$(find cob_bringup)/drivers/sick_visionary_t.launch">
```
fmessmer: add an `unless="$(arg sim)"` for the driver as in https://github.com/ipa320/cob_robots/blob/indigo_dev/cob_bringup/components/cam3d_openni2.launch#L13

cob_bringup/robots/base_solo.launch
```diff
@@ -17,6 +17,7 @@
 	<node pkg="robot_state_publisher" type="state_publisher" name="robot_state_publisher"/>
 
 	<include unless="$(arg use_old_base_drive_chain)" file="$(find cob_bringup)/components/cob4_base.launch">
+                <arg name="robot" value="$(arg robot)"/>
```
fmessmer: use tabs

cob_bringup/robots/cob4-5.xml
```diff
@@ -233,7 +235,7 @@
 				<arg name="robot" value="$(arg robot)"/>
 				<arg name="pc" value="$(arg cob4-5-s1)"/>
 			</include>
-			<include file="$(find cob_bringup)/drivers/sick_visionary_t.launch">
+			<include file="$(find cob_bringup)/components/cam3d_sick_visionary.launch">
 				<arg name="robot" value="$(arg robot)"/>
 				<arg name="name" value="sensorring_cam3d_front"/>
```
fmessmer: pass sim argumetn `<arg name="sim" value="$(arg sim)"/>`

cob_bringup/robots/cob4-5.xml
```diff
@@ -261,7 +263,12 @@
 		</group>
 
 		<include file="$(find cob_bringup)/drivers/mimic.launch">
-			<arg unless="$(arg sim)" name="rotation" value="270"/>
+			<arg unless="$(arg sim)" name="rotation" value="0"/>
```
fmessmer, line 268: `<arg unless="$(arg sim)" name="rotation" value="270"/>` is used for the other robots
does this apply to `cob4-5` as well?

fmessmer (changes requested):
few remarks

fmessmer:
did you test simulation with this branch?

cob_bringup/robots/cob4-5.xml
NadiaHG: No, paul has other configuration

NadiaHG:
@ipa-fxm no, still not tested on simulation

ipa-fxm:
add updated SRDF to this PR

fmessmer (approved):

fmessmer (changes requested):
- add updated SRDF to this PR
 - enable collision_monitor
 - test simulation

NadiaHG:
Already tested on simulation and it is working, just there is a conflict with the coordinate system and topics of the sick_3dcs , has to be discussed https://github.com/ipa320/cob_robots/issues/669 , any way this affects to cob_description (probably the file https://github.com/ipa320/cob_common/blob/indigo_dev/cob_description/urdf/sensors/sick_3dcs.gazebo.xacro) had nothing to do with cob_robots.

The collision monitor will only be enable if it works, I prefer open a separate pull request if https://github.com/ipa320/cob4/issues/480 is solved.

Please merge this pull request if there isn't any other issue with the proposed changes. I would like to clean the robot workspace using the current upstream ipa320/indigo_dev branches.

cob_default_robot_config/robots/cob4-5/script_server/command_gui_buttons.yaml
```diff
@@ -69,7 +69,6 @@ group18:
     [front,move,front],
     [front_down,move,front_down],
     [back,move,back],
-    [back_down,move,back_down],
```
fmessmer, line 72: we also should remove the configs from `cob_default_robot_config/robots/cob4-5/script_server/ 	head_joint_configurations.yaml`

fmessmer (dismissed):
the rest is fine

NadiaHG:
I would like to keep the position, because it is a default position. We remove the button because a move between front_down and back_down could produce a collision between the head and the arms. 

For the arms we are doing the same, we define the default position "ship" but we are not including the button on the command_gui , because move from side or folded to ship in not safe, but the position is needed.

floweisshardt (approved):

floweisshardt:
keeping the position in cob_default_robot_config is fine for me. As long as somebody knows how to prevent collission it's ok to use the default positions. where as non-skilled people should not be able to produce collissions by pressing random buttons on the dashboard

PR #667, opened 2017-05-12
local changes from cob4-7

benmaidel:


cob_bringup/env.sh
```diff
@@ -13,4 +13,6 @@ else
     source /opt/ros/indigo/setup.bash
 fi
 
+export ROS_IP=`hostname -I | awk '{print $1}'`
```
floweisshardt, line 16: @ipa-fxm and @ipa-nhg: do you see any troubles when adding this to our generic `env.sh` script?

cob_hardware_config/robots/cob4-7/config/pc_monitor_cob4-7-h1.yaml
```diff
@@ -2,5 +2,5 @@ check_core_temps: False
 check_ipmi_tool: False
 enforce_clock_speed: False
 num_cores: 4
-load1_threshold: 6
-load5_threshold: 4
+load1_threshold: 8
```
floweisshardt: we should use the same values here as for cob4-2. I'll update this PR

floweisshardt (changes requested):

cob_bringup/robots/cob4-7.xml
```diff
@@ -44,7 +44,6 @@
 		</include>
 		<include file="$(find cob_bringup)/drivers/bms.launch">
 			<arg name="robot" value="$(arg robot)"/>
-			<arg name="sim" value="true"/>
```
fmessmer, line 47: we need this line to be `<arg name="sim" value="$(arg sim)"/>` otherwise we'll break simulation!

fmessmer (changes requested):

cob_bringup/robots/cob4-7.xml
benmaidel: agreed!

cob_bringup/env.sh
fmessmer: what content is in `$1`?
I would like to add this to my the `.bashrc` on my laptop :wink:

cob_bringup/robots/cob4-7.xml
floweisshardt: done in 2147ca8096baa9fac945ec812ada0b67ff3a23a0

cob_bringup/env.sh
benmaidel: `$1` is the first IP Adress which `hostname -I` prints

NadiaHG: In principle should be ok, was already tested on the robot? also sourcing the env.sh of other user to execute the robot.launch?

PR #685, opened 2017-07-19
use cob4-b12 for paul-stuttgart

floweisshardt:


fmessmer:
how about the increased accelerations/decelerations in the local cob4_base_driver.yaml?
still under testing?

fmessmer (approved):

floweisshardt:
Controller updates works although it does not make a big difference. But the drive controllers now produce strange sounds when moving, so we should not commit them as they are at the moment.
Thus I only committed the calibration values.

fmessmer:
> Controller updates works although it does not make a big difference. But the drive controllers now produce strange sounds when moving, so we should not commit them as they are at the moment.
Thus I only committed the calibration values.

so we still have the stuck_detector kicking in sometimes PLUS strange sounds in the motors with the increased accelerations :smile: 
...better to not commit this then :wink:

PR #689, opened 2017-07-21
switch to mimic cpp implementation

benmaidel:
depends on https://github.com/ipa320/cob_driver/pull/344

### DIFF
--- a/cob_bringup/drivers/mimic.launch
+++ b/cob_bringup/drivers/mimic.launch
@@ -1,11 +1,5 @@
 <?xml version="1.0"?>
 <launch>
-
-	<arg name="rotation" default="0"/>
-
-	<!-- start mimic -->
-	<node pkg="cob_mimic" type="mimic_node.py" name="mimic" cwd="node" respawn="false">
-		<param name="rotation" value="$(arg rotation)"/>
-	</node>
-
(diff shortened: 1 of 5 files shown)


floweisshardt (approved):
tested locally: OK

floweisshardt:
why is ```cwd="node"``` neccessary?

cob_bringup/drivers/mimic.launch
```diff
@@ -1,11 +1,5 @@
 <?xml version="1.0"?>
 <launch>
-
-	<arg name="rotation" default="0"/>
-
-	<!-- start mimic -->
-	<node pkg="cob_mimic" type="mimic_node.py" name="mimic" cwd="node" respawn="false">
-		<param name="rotation" value="$(arg rotation)"/>
```
fmessmer, line 8: @ipa-bnm @ipa-fmw @ipa-nhg
just to confirm....there is no need for `rotation` argument on the robots anymore?
I remember @ipa-nhg mentioning that some head displays are configured different than other...on which one did you test?

simulation is fine...

floweisshardt: there's no need for the rotation any more, we'll rotate the ubuntu system settings

PR #702, opened 2017-08-02
Migrate raw3-1 base to canopen

### LINKED ISSUE #422: switch raws and cob3s to new canopen base driver

ipa-fez:
fixes #422 (as it is the last robot to be migrated) :tada: 

See discussion in https://github.com/ipa320/cob_robots/issues/422

Disables stuck detector for raw3-1


jabbenseth (dismissed):
works and looks good

ipa-rmb:
But the torso Power Cube does not work anymore. I will talk to @ipa-bfb and Kevin if we can replace the hardware, soon.

ipa-rmb:
Bruno is not here today and tomorrow, so we have to wait until Monday.

cob_bringup/components/ros_control_base.launch
```diff
@@ -25,10 +26,12 @@
 		<arg name="reset_errors_before_recover" value="$(arg reset_errors_before_recover)"/>
 	</include>
 
-	<include unless="$(arg sim)" file="$(find cob_omni_drive_controller)/launch/stuck_detector.launch">
-		<arg name="timeout" default="2.0"/>
-		<arg name="threshold" default="0.174533"/>
-	</include>
+	<group unless="$(arg disable_stuck_detector)">
```
floweisshardt: why do you need to disable the stuck detector for raw3-1? It's an important feature for the FDMs to not be mechanically damaged in case they're stuck. I don't see a reason why that shouldn't work for raw3 FDMs too.

floweisshardt (changes requested):

cob_bringup/components/ros_control_base.launch
jabbenseth: The stuck detector triggers way too fast for the old ELMOS (-gold) and heavy base (where the rotation unit is and never will be as agile as with the cob4 FDMs). We never had issues without the stuck detector so we removed this new feature. 

We have trouble with the stuck detector on the raw3-3 as well, especially if the batteries are low.

Options are 1) disable 2) fix the stuck detector to work with the old hardware

floweisshardt: you can adjust the parameters 
```
<arg name="timeout" default="2.0"/>
<arg name="threshold" default="0.174533"/>
```

if you set the time to 4sec and the threshold to 0.5 you shouldn't see the stuck detector to get active. If you still see the stuck detector to stop the base there's someting else wrong in the setup (probably controller tuning on elmos).

ipa-rmb:
I talked to Kevin: we have a Schunk Powerball available that can be mounted on raw3-1 and Kevin will be able to do so around 11.08.2017. Mechanically it should not be too complicated. Who could help with the software setup of the new module afterwards? I would be willing to observe but I have no idea how to do it.

cob_bringup/components/raw3_base.launch
```diff
@@ -5,6 +5,7 @@
 
 	<arg name="pkg_hardware_config" default="$(find cob_hardware_config)"/>
 	<arg name="sim" default="false"/>
+	<arg name="disable_stuck_detector" default="false"/>
```
fmessmer: if stuck_detector needs to be disabled (in case @ipa-fmw's https://github.com/ipa320/cob_robots/pull/702/files#r131082757) does not help, I'd prefer positive rather than double-negative: 
`<arg name="enable_stuck_detector" default="true"/>`

fmessmer:
once merged, #422 will be closed...follow-up regarding launch/package cleanup is #703

cob_bringup/components/raw3_base.launch
ipa-fez: Done.

cob_bringup/components/ros_control_base.launch
ipa-fez: @ipa-jba your opinion?

jabbenseth: we could try and pass the "loosened" parameters

cob_bringup/components/ros_control_base.launch
```diff
@@ -25,10 +26,12 @@
 		<arg name="reset_errors_before_recover" value="$(arg reset_errors_before_recover)"/>
 	</include>
 
-	<include unless="$(arg sim)" file="$(find cob_omni_drive_controller)/launch/stuck_detector.launch">
-		<arg name="timeout" default="2.0"/>
-		<arg name="threshold" default="0.174533"/>
-	</include>
+	<group if="$(arg enable_stuck_detector)">
+		<include unless="$(arg sim)" file="$(find cob_omni_drive_controller)/launch/stuck_detector.launch">
+			<arg name="timeout" default="2.0"/>
```
floweisshardt: @ipa-fez: could you successfully use the stuck detector with adapted parameters? if yes, we should remove the parameters from this launch file and move them to a robot specific yaml file in cob_hardware_config and upload that here.

floweisshardt (dismissed):

cob_bringup/components/ros_control_base.launch
ipa-fez: We will test it in the coming days.

fmessmer:
@ipa-jba @ipa-fez
any update regarding the stuck detector? 

I've seen keb working on the robot...
regarding the powerball, I think we can use other Schunk components as a template

also the new arm is mounted...
is this arm already modeled as urdf/xacro?
@ipa-bfb @ipa-rmb

I suggest to finalize the stuck_detector in this pr
and open up a new one (which includes this) for setting up the powerball (and the arm!)

ipa-rmb:
I have no idea whether the F&P arm is already modeled but I would guess so. Bruno was working with it before. He is on holidays this week and can proceed with integrating the arm next week.

ipa-fez:
@ipa-fxm I was going to test yesterday, but there was filming going on or something. I will try to do it this afternoon.

ipa-rmb:
The robot is now available for any software updates or testing. The filming was successfully completed by yesterday lunch time.

benmaidel (approved):

jabbenseth:
is this a solution you are ok with @ipa-fmw?

fmessmer (approved):

bbrito:
@ipa-fxm the full robot is already modelled and as urdf/xacro. You may find it here:

https://github.com/ipa-bfb/p_rob.git

PR #712, opened 2017-08-24
single computer for raw, fix ports

jabbenseth:
Migrate raw3-3 to kinetic. Because now only one pc is used for all the "hardware" stuff the interface names changed for "light" and we use "ttyScanX" because it has read and write permissions for all users thanks to [99-cob.rules](https://github.com/ipa320/setup/blob/master/udev_rules/99-cob.rules)

### DIFF
--- a/cob_bringup/tools/autoinit.launch
+++ b/cob_bringup/tools/autoinit.launch
@@ -1,0 +1,10 @@
+<?xml version="1.0"?>
+<launch>
+	<node pkg="cob_helper_tools" type="auto_init.py" name="auto_init" output="screen">
+		<rosparam param="components">[base]</rosparam>
+	</node>
+
+	<node pkg="cob_helper_tools" type="auto_recover.py" name="auto_recover" output="screen">
+		<rosparam param="components">[base]</rosparam>
+	</node>
+</launch>
--- a/cob_hardware_config/robots/raw3-3/config/light.yaml
+++ b/cob_hardware_config/robots/raw3-3/config/light.yaml
@@ -1,5 +1,5 @@
 devicedriver: ms-35
-devicestring: /dev/ttyUSB2
+devicestring: /dev/ttyLED
 baudrate: 38400
 invert_output: false
 pubmarker: false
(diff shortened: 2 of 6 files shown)


ipa-fez (dismissed):

stefandoerr:
@ipa-bnm please review

cob_hardware_config/robots/raw3-3/config/light.yaml
```diff
@@ -1,5 +1,5 @@
 devicedriver: ms-35
-devicestring: /dev/ttyUSB0
+devicestring: /dev/ttyUSB2
```
benmaidel: We should add a udev rule on raw3-3 for the led's. Then we can name the device `ttyLED`.

benmaidel (dismissed):

stefandoerr:
> We should add a udev rule on raw3-3 for the led's. Then we can name the device ttyLED.

In this PR or can this be merged?

ipa-fez:
~Please wait for https://github.com/ipa-jba/cob_robots/pull/2~

edit @ipa-jba: changes implemented in 540b607 and 1d6f49d

ipa-fez (approved):

benmaidel (approved):

cob_bringup/tools/autoinit.launch
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0"?>
+<launch>
+	<node pkg="cob_helper_tools" type="auto_init.py" name="auto_init" output="screen">
+		<rosparam param="components">[base]</rosparam>
```
fmessmer, line 4: the `components` should be parameterizable in order to be usable for other robots with e.g. components `[base, torso, head]`, i.e. cob4

in msh, we introduced a new [`components.yaml`](https://github.com/ipa320/msh/blob/indigo_dev/msh_robot_config/robots/cob4-7/config/components.yaml) and [use it like so](https://github.com/ipa320/msh/blob/indigo_dev/msh_bringup/launch/application.xml#L59-L64)

we could discuss about moving this into yaml into each robots bringup layer....i.e. from `msh` to `cob_robots`
@ipa-fez @ipa-jba what do you think?

fmessmer (commented):
post-merge-review

cob_bringup/tools/autoinit.launch
ipa-fez: AFAIK we only use auto-bringup on the raw3-3 (which I made this file for) but It sounds nice to have it in a generic way.

PR #719, opened 2017-09-10
anon machine tags

fmessmer:
needs to be tested on robot for node deployment verification!


floweisshardt:
why do we need anon machine tags?

fmessmer:
> why do we need anon machine tags?

there were issues in the roslaunch checks when testing `cob_bringup` robots and `unity_bringup` robots alongside each other

fmessmer:
works as expected

floweisshardt (approved):

PR #728, opened 2017-10-11
setup cob4-7

NadiaHG:
- added arms configuration
 - commented head and sensorring

### DIFF
--- a/cob_hardware_config/robots/cob4-7/urdf/cob4-7.urdf.xacro
+++ b/cob_hardware_config/robots/cob4-7/urdf/cob4-7.urdf.xacro
@@ -65,7 +65,7 @@
     <origin xyz="${torso_cam3d_down_x} ${torso_cam3d_down_y} ${torso_cam3d_down_z}" rpy="${torso_cam3d_down_roll} ${torso_cam3d_down_pitch} ${torso_cam3d_down_yaw}" />
   </xacro:realsense>
 
-  <xacro:head name="head" parent="torso_3_link" dof1="true" dof2="true" dof3="true">
+  <xacro:head name="head" parent="torso_3_link" dof1="false" dof2="false" dof3="false"> <!-- HotFix: set dof* to true when the HW is repaired -->
     <origin xyz="${head_x} ${head_y} ${head_z}" rpy="${head_roll} ${head_pitch} ${head_yaw}" />
   </xacro:head>
   <xacro:usb_cam name="head_cam" parent="head_3_link" ros_topic="head_cam" update_rate="10" height="960" width="1280" > <!-- get this from config/head_cam.yaml -->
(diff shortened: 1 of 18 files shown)


NadiaHG:
this PR requires https://github.com/ipa320/cob_calibration_data/pull/142
@cagbal @ipa-rmb FYI

cob_hardware_config/robots/cob4-7/urdf/cob4-7.urdf.xacro
```diff
@@ -15,6 +15,12 @@
   <!-- torso -->
   <xacro:include filename="$(find cob_description)/urdf/cob4_torso/torso.urdf.xacro" />
 
+  <!-- arm -->
+  <xacro:include filename="$(find cob_description)/urdf/cob4_arm/cob4_arm.urdf.xacro" />
+
+  <!-- hand -->
+  <xacro:include filename="$(find cob_description)/urdf/cob4_gripper/gripper.urdf.xacro" />
```
fmessmer, line 22: grippers are not added to the urdf...and also not part of the robot.xml...is that intended?

cob_hardware_config/robots/cob4-7/urdf/cob4-7.urdf.xacro
```diff
@@ -66,6 +72,14 @@
     <origin xyz="${head_cam_x} ${head_cam_y} ${head_cam_z}" rpy="${head_cam_roll} ${head_cam_pitch} ${head_cam_yaw}" />
   </xacro:usb_cam>
```
fmessmer, line 73: as long as head is disabled in robot.xml, you might need to set the dof's to false...otherwise it will collapse in simulation
...same for sensorring!

cob_hardware_config/robots/cob4-7/urdf/cob4-7.urdf.xacro
NadiaHG: the motors of the hands are missed/broken... I am not sure if the hands will be replaced,  still to be clarified...

cob_hardware_config/robots/cob4-7/urdf/cob4-7.urdf.xacro
NadiaHG: again components broken, to be clarified.. but if the sim will not  work I can temporally update the URDF

fmessmer: :+1:
for explanation:
in simulation the head will hang loose because there is no controller because they are not loaded because the head part is commented in the robot.xml
setting the dof's to false makes the joints fixed instead of revolute and thus no controllers are needed

fmessmer:
@ipa-rmb @ipa-bfb @cagbal
please review - it won't get merged otherwise

@ipa-jba @ipa-fez

fmessmer (approved):

ipa-rmb (approved):
To me it looks fine, but what can I tell with my limited arm/hardware/config knowledge and experience. Maybe @ipa-bfb can give a better assessment.

By the way, if the robot is for 324 then I would expect that every component works (is this the same robot as referred to in Johannes' mail?) - or is there some other agreement by our bosses?

cagbal:
I could not understand that Johannes mentioned a cob4-2 robot, but this is about cob4-7. Is this the robot which will stay with us?

fmessmer:
you need to get to know your robots :wink:

PR #737, opened 2017-11-07
changed group name/joints for raw3-1

ipa-rmb:
- for grasping objects with the gripper, necessary for moveit planning
- successfully tested on raw3-1


ipa-rmb:
Replaced/included by/in #758

PR #751, opened 2018-02-22
[WIP] test position twist controller

fmessmer:
testing https://github.com/ipa320/cob_control/pull/140

**should not be merged like this as it is not dual-distro compatible**

### DIFF
--- a/cob_hardware_config/robots/common/cob4_base_controller.yaml
+++ b/cob_hardware_config/robots/common/cob4_base_controller.yaml
@@ -8,8 +8,12 @@ joint_state_controller:
 
 ## twist_controller
 twist_controller:
-  type: cob_omni_drive_controller/WheelController
+  type: cob_omni_drive_controller/WheelMultiController
   required_drive_mode: 3
+  required_drive_modes:
+    fl_caster_rotation_joint: 1
+    b_caster_rotation_joint: 1
+    fr_caster_rotation_joint: 1
   #max_rot_velocity: 1.0
   #max_trans_velocity: 1.0
   pub_divider: 1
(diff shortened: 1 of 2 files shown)


fmessmer:
I tested this together with @ipa-tif on cob4-18

we found `10000000` to be the best acceleration limit for the steer_joints (higher acceleration limits result in over-shoot and doesn't result in a faster conversion against the command) - tested with single position joint controller and step command publisher via `rqt`

using this setting with the position twist_controller however showed some effects that are not yet ideal:
 - controller wheel_commands sometimes show sign switches 1.5->-1.5->1.5 within subsequent control steps - might be some shortest angular distance issue
 - also the [twist_controller parameter](https://github.com/ipa-fxm/cob_robots/blob/test_position_twist_controller/cob_hardware_config/robots/common/cob4_base_controller.yaml#L24-L28) might have an effect on the wheel_commands itself :point_right: tuning
 - unfortunately, the position twist_controller aka. wheel_multi_controller currently does not support dynamic reconfigure for those values - porting this from the velocity twist_controller aka. wheel_controller might help with the tuning

@ipa-bnm @ipa-mdl 
FYI

mathias-luedtke:
> also the twist_controller parameter might have an effect on the wheel_commands itself point_right tuning

The controller params do not have any effect.
The command is just the direction of the input velocity.

> controller wheel_commands sometimes show sign switches 1.5->-1.5->1.5 within subsequent control steps - might be some shortest angular distance issue

It's a little bit more [complicated](https://github.com/ipa320/cob_control/blob/kinetic_dev/cob_omni_drive_controller/src/UndercarriageCtrlGeom.cpp#L104). Not sure if the current behavior was ported correctly and if it makes sense at all.

fmessmer:
won't merge for `cob_robots`

Issue #732, opened 2017-10-24
use laser_filter pipeline for scanner


fmessmer:
according to @ipa-jba, [laser_filters](http://wiki.ros.org/laser_filters) is quite powerful

investigate setting up a proper filter_pipeline for e.g. the sick scanners that would replace:
 - [cob_scan_filter](https://github.com/ipa320/cob_robots/blob/indigo_dev/cob_bringup/drivers/laser_scan_filter.launch) via http://wiki.ros.org/laser_filters#LaserScanAngularBoundsFilter
 - ~[max_range_fix](https://github.com/ipa320/cob_driver/pull/358) via http://wiki.ros.org/laser_filters#LaserScanRangeFilter~

@ipa-bnm @ipa-fmw FYI

fmessmer:
@ipa-srd
can you explain again, why it was necessary to avoid `NaN` in the laser scan and replace it with `max_range`?

stefandoerr:
With https://github.com/ipa320/cob_driver/pull/358 max range measurements are properly set, i.e. their values are >= than the configured max_range of the respective laser scanner. This is needed so that a node can properly identify max range measurements within the ranges array. For example. the localization module normally filters out max range measurements to avoid matching those measurements against the map. 
However, in the current setup, this would lead to an undesired behavior in the costmap which will discard those max range measurements for ray tracing. In open areas where you have a lot of max range measurements, this leads to the problem that dynamic objects are not cleared properly. With the feature implemented by @ipa-jba in https://github.com/ros-perception/laser_filters/pull/56, this problem can be resolved. The filter sets all max range measurements to inf. Together with setting the `inf_is_valid` param in the obstacle_layer to true, this fixes the problem.

stefandoerr:
Example config:
```
scan_filter_chain:
 - name: self_meas
   type: LaserScanAngularBoundsFilter
   params:
     lower_angle: 0.78
     upper_angle: 3.91
 - name: max_range
   type: LaserScanRangeFilter
   params:
     use_message_range_limits: true
upper_replacement_value: .inf

```
Example node

```
+  <node pkg="laser_filters" type="scan_to_scan_filter_chain" name="laser_filter">
+    <rosparam command="load" file="$(find striml_hardware_config)/robots/$(arg robot)/config/laser_filter_chain.yaml" />
+    <remap from="scan" to="/base_laser_front/scan" />
+    <remap from="scan_filtered" to="scan" />
+  </node>
```

[...]

PR #757, opened 2018-04-27
simplify config structure

fmessmer:
this is a backport of some simplifications/unifications/parameter_updates we did in `unity_robots` which improve daily robot performance and also simplifies robot configuration (mostly by using symlinks rather than needing to copy full files)

it will also bring in a few new features such as the `scan_unified_ranged` discussed in #732


fmessmer:
This is good to merge now!

I verified all my changes by comparing **before** and **after** using [`roslaunch-dump`](https://gist.github.com/ipa-mdl/66b14a895e13003a2f76e483bd00aa63) by @ipa-mdl 
I did this to the best of my belief!

However, it is made over-complicated by old hardware and/or not sticking to the scheme when a new robot was set up in the past...in particular, `flexisoft` settings are a mess!
I resolved this by now considering our latest robots `cob4-13` and `cob4-16` to be the standard for future robots...those configs were used as templates/targets for the symlinks...
In case an old robot is not conform with the new standard, such robot keeps his own copy of a config file rather than using a symlink...

fmessmer:
I think testing this on `cob4-7` (Paul Stuttgart) and `cob4-13` (Cardiff - not yet shipped) should be good enough as a final test...
@HannesBachter could you help here?

Maybe @ipa320/navigation-push could try this on `cob4-18` (323)

fmessmer:
~~ToDo: fix roslaunch-checks for `cob_default_robot_config`~~
done

fmessmer:
tested on cob4-7
unfortunately, cob4-13 is shipped already
~~I will do one more test with cob4-10 (which recently got a hw update - not yet proposed as a pr yet)~~
cob4-10 has not disk space anymore...thus, is currently not operable...

fmessmer:
I did these changes to the best of my knowledge...always double-checking the before-after roslaunch-dump results for each robot...

I cannot guarantee that minor yaml adjustments might need to be made by one or the other robot...where the real hardware differs from the standard...but as I do not have access to all those robots...this needs to be done when those robots are updated the next time...e.g. `cob4-10`, `raw3-1`

as no one ever reviews prs in this repo except me and TravisCI, I'm going to merge

PR #758, opened 2018-05-03
added updated configuration files of raw3-1

ipa-rmb:
This is the current configuration of raw3-1. There are no other changes compared to current cob_robots/indigo_dev. I hope you do not complain about the stomp configuration files. They are just text and do not harm anything and do not introduce any dependencies. But if we do not check them in, this would be the only deviation from 320 cob_robots. So checking all in would be much more convenient for us.


fmessmer (commented):
although these files require `stomp` as a dependency, I don't care whether this is merged or not, as these yaml files are nowhere used from within this repo....thus, they do not break any TravisCI or release...

PR #759, opened 2018-05-08
re-add joint_names param + consistent launch structure for sdhx with remote driver

fmessmer:
@HannesBachter 
this should fix the problem at UH


fmessmer:
I decided to revert the recent changes in `cob_command_tools` which caused more trouble than it simplified things (i.e. reduce duplicate params) - see https://github.com/ipa320/cob_command_tools/pull/220

thus, I re-added the `joint_names` param back for all components in `cob_default_robot_config` in https://github.com/ipa320/cob_robots/pull/759/commits/8accd4060cdecbd4ba25ea93d76b6602af56d444
in addition, I added a launch file for the grippers where the driver is started remotly - which fixes the simulation for those robots....it's also more consistent to the sdhx_legacy concept!

fmessmer:
we were able to test this on `cob4-10` see #760 

@pholthau FYI

PR #760, opened 2018-05-08
cob4-10 hw upgrade

fmessmer:
prerequisite for https://github.com/ipa320/cob4/issues/669

current issues are documented in https://github.com/ipa320/cob4/issues/669#issuecomment-387453881
let's keep the discussion there


cob_hardware_config/robots/cob4-10/config/light_torso.yaml
```diff
@@ -1 +0,0 @@
-../../cob4/config/light_torso.yaml
```
fmessmer, line 1: this only reverts what has unintendedly been modified in #757 
i.e. the `cob4-10/light_torso.yaml` has been different from the common file...thus, the symlink is reverted here

fmessmer:
tested on HW...

PR #775, opened 2019-03-21
add cob4-25

fmessmer:
new research robot

ref. ipa320/cob_supported_robots#19
ref. ipa320/cob_calibration_data#150


cob_hardware_config/robots/cob4-25/config/base_driver.yaml
```diff
@@ -0,0 +1,12 @@
+nodes:
+  fl_caster_rotation_joint:
```
floweisshardt, line 2: same values as in https://github.com/mojin-robotics/mojin_calibration_data/pull/44

PR #800, opened 2020-02-20
[DO NOT MERGE] Backup/kevin cob4 18

floweisshardt:
backup from @ipa340 Kevin from cob4-18

PR #802, opened 2020-02-20
add autorecover for kevin - cob4-20 + cob4-23

floweisshardt:


fmessmer (changes requested):
this does not update cob4-20 but cob4-18...is this correct?
@floweisshardt

floweisshardt:
> this does not update cob4-20 but cob4-18...is this correct?

no sorry, this is wrong. We should not touch cob4-18 (@ipa320) but add the autorecover to cob4-20 and cob4-23 (both @ipa340 as Kevins).

fmessmer (changes requested):
possibly include backup from #800 - once https://github.com/ipa320/cob_robots/pull/800#issuecomment-591330787 is resolved

do they also want `auto_init`?
because then we should just add https://github.com/ipa320/cob_robots/blob/kinetic_dev/cob_bringup/tools/autoinit.launch

cob_bringup/robots/cob4-18.xml
```diff
@@ -57,6 +57,9 @@
 			<arg name="robot" value="$(arg robot)"/>
 			<arg name="sim" value="$(arg sim)"/>
 		</include>
+		<node pkg="cob_helper_tools" type="auto_recover.py"  name="auto_recover_base" output="screen">
```
floweisshardt, line 60: remove

floweisshardt:
not needed anymore. the consolidated version is https://github.com/ipa320/cob_robots/pull/804

PR #810, opened 2020-07-15
canTransform argument target_frame in tf2 frame_ids cannot start with…

ipa-fog:
canTransform argument target_frame in tf2 frame_ids cannot start with a '/' . The terminal indicates this by a lot of warnings.

- fixed by removing all '/' prefix from frames.
- tested with melodic in simulation


fmessmer:
seems legit - @ipa-fog can you give an example for a node that throws warnings without this fix, so i can verify this?

fmessmer (approved):

PR #817, opened 2021-05-03
NODE testing config

fmessmer:
ref https://github.com/mojin-robotics/cob4/issues/1758

provides a simple cob4 base with two additional cameras (name and origin as on mojin_robotics' LUKA) for navigation tests in simulation


fmessmer:
has successfully been used by NODE for testing

PR #831, opened 2023-03-02
add moveit config and launch files for cob4-30

FaycalNait:
added a moveit config for `arm_left` and `arm_right` with the MoveitSetupAssistant wizard for cob4-30


fmessmer:
sorry, the `cob_moveit_config` package is not meant to be used like this - it goes together with `cob_manipulation/cob_moveit_bringup`!
but these packages are no longer actively maintained!

please use the `moveit_setup_assistant` to create your own, cob4-30 specific moveit package which contains launch and config files!
you will have to host it on github yourself - ipa320 is only meant for the bringup layer of the robot - moveit is not part of that...

